Issue #3, opened 2025-01-07
Add Metric for Bundle Expiration


aaronbrethorst:
# user story

As a transit manager,
I want to see how many days remain until my bundle expires,
so I can remember to update it 

# acceptance criteria 

- [x] add a static GTFS parsing library: https://github.com/jamespfennell/gtfs
- [x] download the static GTFS once per day and cache it for subsequent metrics reporting
- [x] add a new metrics file/function at `internal/metrics/bundle_expiration.go` / `CheckBundleExpiration()`, and wire it up in `api/metrics.go`
- [x] Push days until bundle expiration into Prometheus (with a gauge status, I think)
- [x] Some tests verifying that this works as expected

PR #5, opened 2025-01-12
Feature/gtfs-expiration-metrics

Ahmedhossamdev:
fixes: #3


cmd/api/main.go
```diff
@@ -52,6 +54,29 @@ func main() {
 
 	app.startMetricsCollection()
 
+	// Download the GTFS bundle on startup
+	cachePath := "cache/gtfs.zip"
```
aaronbrethorst, line 58: I'd like to see the cache path be unique on a per-URL basis. How about the URL gets passed in as a command line option and the cache path becomes `"cache/{SHA 1 value of the URL}.zip"

aaronbrethorst (changes requested):
looking really good! please see my one command and change accordingly. 🚀

cmd/api/main.go
Ahmedhossamdev: Done

aaronbrethorst (approved):
🔥 great work

Issue #19, opened 2025-01-21
Update README with information about config loading


aaronbrethorst:
Issue #14 changed how configuration data is loaded into Watchdog, but the README hasn't been updated. It should be updated with information about both loading a local JSON file and a remote URL (including authentication).

Issue #20, opened 2025-01-21
Update Dockerfile with remote URL configuration support


aaronbrethorst:
The Dockerfile should be updated to load a remote URL (complete with support for authentication).

PR #25, opened 2025-01-21
Update/readme-dockerfile-config

Ahmedhossamdev:
fixes: #19 #20

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,0 @@
-# .env.example
-
-FEED_API_KEY="your_api_key_here"
-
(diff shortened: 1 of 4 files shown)


aaronbrethorst (changes requested):
This is looking good, but I want to see the Dockerfile changed such that we don't need to customize or override its behavior for it to work. Check out how Amr approached this for Prometheus: https://github.com/OneBusAway/obacloud-monitoring/tree/main/prometheus

aaronbrethorst (approved):
great work 👍

PR #43, opened 2025-02-15
add ScheduleRouteForStop for metrics and add method for getting the l…

jerryyummy:
add metrics for get all routes for one stop and we can monitor all the routes across the stop
get the location for a stop


jerryyummy:
also add metric to predict arrival rate

Issue #28, opened 2025-01-24
New Metric: Scheduled Trips


Ahmedhossamdev:
#### **User Story**:
As a DevOps engineer,  
I want to track the number of scheduled trips for each agency and route (we can make for routes first)
so that I can compare scheduled data with real-time updates.

#### **Acceptance Criteria**:
- Add a new metric to count scheduled trips for each agency and route.
- Add tests to verify the metric is working correctly.

PR #48, opened 2025-04-01
New Metric: Scheduled trips for route and agency

HarshKh13:
for issue #28


HarshKh13:
@aaronbrethorst I have added the functionality to get all the scheduled trips for a route and agency. Will be adding the test cases soon.

amrhossamdev (changes requested):
Great job, @HarshKh13!

I need you to write a test for the 'scheduled trips' service.

Additionally, please update the `startMetricsCollection()` function in `cmd/watchdog/metrics.go` to integrate the `scheduled trips` service. This will enable us to gather data for each server and ensure the service is fetched when we call `app.startMetricsCollection()`.

Let me know if you'd like further tweaks!

Issue #56, opened 2025-05-02
Add support for the REST API server's new Metrics API


aaronbrethorst:
OBA REST API 2.6.0 has a new `metrics` API that exposes data that should be consumed by the Watchdog project and made available to Prometheus. Check out https://api.unitrans.onebusawaycloud.com/api/where/metrics.json?key=org.onebusaway.iphone for a live example of one of these feeds.

Each of the data points in this API can be matched up to a metric exposed by the original watchdog project, described here: https://github.com/OneBusAway/onebusaway-application-modules/wiki/OneBusAway-Watchdog#metrics

# Acceptance criteria 

- [ ] Load metrics.json data from all configured OBA servers
- [ ] Gracefully handle case where an OBA server doesn't have a metrics.json endpoint (i.e. pre-2.6.0 servers)
- [ ] expose new metrics to Prometheus for each datapoint in the metrics API

PR #57, opened 2025-05-03
Add OBA API metrics collection and Prometheus gauges

amrhossamdev:
About #56


internal/metrics/oba_rest_api_metrics.go
```diff
@@ -0,0 +1,103 @@
...
+			StopIDsMatchedCount         map[string]int      `json:"stopIDsMatchedCount"`
+			StopIDsUnmatched            map[string][]string `json:"stopIDsUnmatched"`
+			StopIDsUnmatchedCount       map[string]int      `json:"stopIDsUnmatchedCount"`
+			TimeSinceLastRealtimeUpdate map[string]int      `json:"timeSinceLastRealtimeUpdate"`
+		} `json:"entry"`
+	} `json:"data"`
+}
+
+const (
+	baseURLTemplate = "https://api.%s.onebusawaycloud.com/api/where/metrics.json"
```
aaronbrethorst, line 33: there's no guarantee this server will have a onebusawaycloud.com domain name. Instead, use `ObaBaseURL` from the Server model.

internal/metrics/oba_rest_api_metrics_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"gopkg.in/dnaeon/go-vcr.v4/pkg/recorder"
+	"net/http"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+func TestFetchObaAPIMetrics_WithVCR(t *testing.T) {
```
aaronbrethorst, line 11: this looks good. I'd like to see a negative test, too. i.e. using api.pugetsound.onebusaway.org, which doesn't support the metrics API

aaronbrethorst (changes requested):
looking good. I see a couple changes to make before merging!

internal/metrics/oba_rest_api_metrics_test.go
amrhossamdev: I did that here 

``` GO
		err := FetchObaAPIMetrics(tt.slugID, tt.serverUrl, tt.apiKey, client)

			if tt.wantErr {
				if err == nil {
					t.Error("expected error but got none")
					return
				}
				return
			}

			if err != nil {
				t.Errorf("unexpected error: %v", err)
			}
```

internal/metrics/oba_rest_api_metrics.go
amrhossamdev: Done

aaronbrethorst (approved):
nice work!

PR #59, opened 2025-06-09
Clarify config usage: rename config.json to config.json.template and ignore actual config.json

Abo-Omar-74:
- Renamed `config.json` to `config.json.template` to explicitly communicate that the file contains example configuration data meant to be copied and customized by users.
- Stopped tracking `config.json` in Git by removing it from the index to prevent accidental commits of sensitive or local configuration.
- Confirmed `config.json` is included in `.gitignore` to avoid future tracking by contributors.
- Updated documentation with clear instructions on how to copy `config.json.template` to `config.json` and customize it for local use.


aaronbrethorst (approved):
Looks good!

PR #67, opened 2025-06-26
test: Make Integration Config Path Configurable via Test Flag

Abo-Omar-74:
* Introduced configuring url parsing logic using flags and move loading logic to `integration_test.go`
* Refactored `TestDownloadGTFSBundles` and `TestOBAConnection` to use shared config setup
* Updated README with instructions for running integration tests using `-integration-config` flag


aaronbrethorst (approved):
👍 looks good

PR #74, opened 2025-07-12
feat: Add vehicle telemetry tracking with computed speed and discrepancy metrics

Abo-Omar-74:
* Renamed `TrackVehicleReportingFrequency` → `TrackVehicleTelemetry` to reflect expanded functionality.
* Introduced `LastSeen` struct to store vehicle timestamp and coordinates.
* Calculated speed using Haversine distance between consecutive GTFS-RT reports.
* Added Prometheus gauge: `gtfs_rt_vehicle_computed_speed`.
* Added Prometheus gauge: `gtfs_rt_vehicle_speed_discrepancy_ratio` to detect anomalies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.23.5
 require (
 	github.com/OneBusAway/go-sdk v0.1.0-alpha.13
 	github.com/getsentry/sentry-go v0.33.0
-	github.com/golang/geo v0.0.0-20250630213057-fac2d31592dd
+	github.com/golang/geo v0.0.0-20250707181242-c5087ca84cf4
 	github.com/jamespfennell/gtfs v0.1.24
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/prometheus/client_golang v1.20.5
--- a/go.sum
+++ b/go.sum
@@ -11,8 +11,8 @@ github.com/getsentry/sentry-go v0.33.0 h1:YWyDii0KGVov3xOaamOnF0mjOrqSjBqwv48UEz
 github.com/getsentry/sentry-go v0.33.0/go.mod h1:C55omcY9ChRQIUcVcGcs+Zdy4ZpQGvNJ7JYHIoSWOtE=
 github.com/go-errors/errors v1.4.2 h1:J6MZopCL4uSllY1OfXM374weqZFFItUbrImctkmUxIA=
 github.com/go-errors/errors v1.4.2/go.mod h1:sIVyrIiJhuEF+Pj9Ebtd6P/rEYROXFi3BopGUQ5a5Og=
-github.com/golang/geo v0.0.0-20250630213057-fac2d31592dd h1:4BU/Fy2KwDucgY0al91L0wJI05rb1Y4hQKucQkybjT8=
-github.com/golang/geo v0.0.0-20250630213057-fac2d31592dd/go.mod h1:Vaw7L5b+xa3Rj4/pRtrQkymn3lSBRB/NAEdbF9YEVLA=
+github.com/golang/geo v0.0.0-20250707181242-c5087ca84cf4 h1:vCeHcs8N7MOccOOsOVIy1xcYu+kBkA4J5urTgigww7c=
+github.com/golang/geo v0.0.0-20250707181242-c5087ca84cf4/go.mod h1:AN0OjM34c3PbjAsX+QNma1nYtJtRxl+s9MZNV7S+efw=
 github.com/google/go-cmp v0.7.0 h1:wk8382ETsv4JYUZwIsn6YpYiWiBsYLSJiTsyBybVuN8=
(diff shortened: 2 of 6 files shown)


Abo-Omar-74:
**How do I calculate the speed of a vehicle and evaluate its accuracy against GTFS-reported speed?**
For each vehicle, I keep track of its previous coordinate and timestamp. When a new API response arrives, I compute the distance traveled using a Haversine-based method and divide it by the time difference to get the estimated speed. I then compare this value to the GTFS-reported speed to quantify any discrepancy.

**What is the method used to calculate this distance?**
I implemented a `HaversineDistance` function that uses the equation: `d = R * c`

Where `d` is the distance between two geographic points, `R` is the Earth's radius, and `c` is the angular distance in radians. The angular distance is obtained using the `Distance()` method from the `golang/geo` package, which returns the angle between two `LatLng` points using the Haversine formula internally. The full derivation of this formula is documented [here](https://www.movable-type.co.uk/scripts/latlong.html)

**Is this approach standard or reliable?**
Yes, this same method, computing the angular distance first, then scaling it by Earth's radius, is also used in other libraries like [goigc](https://pkg.go.dev/github.com/ezgliding/goigc/pkg/igc#Point.Distance). You can view its implementation [here](https://github.com/ezgliding/goigc/blob/v0.4.0/pkg/igc/point.go#L153). However, since `goigc` hasn't been maintained since 2020, I preferred a self-contained implementation that builds on top of the actively maintained `golang/geo`.

**How accurate is the Haversine approach, and is it suitable for this use case?**
The Haversine formula, as described on [Movable Type Scripts](https://www.movable-type.co.uk/scripts/latlong.html), assumes a spherical Earth. This simplification introduces a small margin of error, typically up to 0.3%. The spherical model works well in our case, as we're identifying significant speed mismatches, not performing sub-meter geolocation.

**Why define the threshold in PromQL/Grafana rather than hardcoding it in Go?**
Placing the speed discrepancy threshold in PromQL or Grafana provides greater flexibility; it can be adjusted instantly, supports alerting mechanisms natively, and avoids redeploying the Go service just to tweak a config value. This makes the system more responsive and easier to maintain.

Abo-Omar-74:
**How do I calculate the speed of a vehicle and evaluate its accuracy against GTFS-reported speed?**
For each vehicle, I keep track of its previous coordinate and timestamp. When a new API response arrives, I compute the distance traveled using a Haversine-based method and divide it by the time difference to get the estimated speed. I then compare this value to the GTFS-reported speed to quantify any discrepancy.

**What is the method used to calculate this distance?**
I implemented a `HaversineDistance` function that uses the equation: `d = R * c`

Where `d` is the distance between two geographic points, `R` is the Earth's radius, and `c` is the angular distance in radians. The angular distance is obtained using the `Distance()` method from the `golang/geo` package, which returns the angle between two `LatLng` points using the Haversine formula internally. The full derivation of this formula is documented [here](https://www.movable-type.co.uk/scripts/latlong.html), and this method of calculation is also discussed in [this issue](https://github.com/golang/geo/issues/63) in the official golang/geo repository

**Is this approach standard or reliable?**
Yes, this same method, computing the angular distance first, then scaling it by Earth's radius, is also used in other libraries like [goigc](https://pkg.go.dev/github.com/ezgliding/goigc/pkg/igc#Point.Distance). You can view its implementation [here](https://github.com/ezgliding/goigc/blob/v0.4.0/pkg/igc/point.go#L153). However, since `goigc` hasn't been maintained since 2020, I preferred a self-contained implementation that builds on top of the actively maintained `golang/geo`.

**How accurate is the Haversine approach, and is it suitable for this use case?**
The Haversine formula, as described on [Movable Type Scripts](https://www.movable-type.co.uk/scripts/latlong.html), assumes a spherical Earth. This simplification introduces a small margin of error, typically up to 0.3%. The spherical model works well in our case, as we're identifying significant speed mismatches, not performing sub-meter geolocation.

**Why define the threshold in PromQL/Grafana rather than hardcoding it in Go?**
Placing the speed discrepancy threshold in PromQL or Grafana provides greater flexibility; it can be adjusted instantly, supports alerting mechanisms natively, and avoids redeploying the Go service just to tweak a config value. This makes the system more responsive and easier to maintain.

aaronbrethorst (approved):
looks good! at some point it would be good to follow up on the copilot feedback, but that can happen later.

Abo-Omar-74:
> looks good! at some point it would be good to follow up on the copilot feedback, but that can happen later.

Sorry, I didn’t apply the Copilot suggestion earlier. I was sick over the past few days and overlooked the changes. However, I’ve applied the suggestion in the #75 PR.

Regarding the threshold, I intentionally chose to report all distance values to Prometheus and defer thresholding to the PromQL/Grafana layer, so the threshold can be adjusted more easily without requiring a redeploy.

That said, Copilot’s suggestion makes sense to avoid reporting very small distance differences that are likely just GPS noise.

I’ll look into the typical accuracy of the reported GTFS-RT coordinates to determine a reasonable threshold. I plan to include this in a future PR after discussing it with you.

PR #75, opened 2025-07-15
Encapsulate Config Access with Internal Locking and Thread-Safe Getters

Abo-Omar-74:
* Moved `UpdateConfig` and `sync.RWMutex` from `Application` to `Config` for better encapsulation
* Added `GetServers()` method with read-lock for safe concurrent access
* Replaced all direct access to `Config.Servers` with `GetServers()` or `UpdateConfig()`
* Updated `GetServers()` to return a copy to prevent shared mutation and ensure thread safety

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 require (
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
-	github.com/golang/geo v0.0.0-20250707181242-c5087ca84cf4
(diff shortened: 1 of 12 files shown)


aaronbrethorst:
needs a rebase, now.

Abo-Omar-74:
@aaronbrethorst 
I’ve rebased the branch and applied the requested changes.

Abo-Omar-74:
> looking good! copilot had some feedback, but otherwise we're ready to go with this.

I've applied copilot suggestions.

PR #76, opened 2025-07-19
feat: enhance telemetry with concurrency-safe vehicle tracking, graceful shutdown, and Prometheus metrics

0xaboomar:
### Metrics & Telemetry Enhancements

- **Encapsulate telemetry data in a concurrency-safe structure**
    - Wrap `map[int]map[string]LastSeen` inside a `VehicleLastSeen` struct with `sync.RWMutex`
    - Provide thread-safe utility methods (`Get`, `Set`, `Count`) for access and mutation
    - Add `ClearRoutine` to periodically remove stale vehicle data based on a time threshold
* **Inject `VehicleLastSeen` into `Application` struct**
* **Add application-wide context**
  * Use `context.WithCancel` in `main` for graceful shutdown
  * Enables future routines (like data cleanup) to terminate cleanly
* **Add Prometheus metric for currently tracked vehicles**
  * Expose `gtfs_rt_tracked_vehicles_count` as a gauge
  * Add `Count(serverID)` to get active vehicle count
---

### Utilities & Time Handling

* **Add `CustomTime` type for JSON marshaling of `time.Time`**

  * Supports `YYYYMMDD` format
  * Implements `MarshalJSON` and `UnmarshalJSON`
  * Provides `.Time()` helper to convert back to `time.Time`

* **Add generic helpers to persist map as JSON**

  * `SaveMapToFile` writes a map to a JSON file
  * `LoadMapFromFile` reads a map from a JSON file
---
### Refactoring 
* **Refactor service date range logic**
  * Moved earliest/latest service date calculation into the `gtfs` package
  * Centralizes logic for reuse (e.g., in `CheckBundleExpiration`)


Abo-Omar-74:
## Follow Up 

Added a context with cancellation for all background goroutines.

why? 

Currently, background goroutines (metrics collection, GTFS refresh, vehicle cleanup) do not shut down gracefully when the server stops. This can lead to resource leaks or incomplete cleanup.

aaronbrethorst (changes requested):
this will be ready to go once Copilot's feedback is addressed. great work.

PR #77, opened 2025-07-26
feat: enable connection pooling via shared http.Client and reuse RealtimeStore across GTFS functions

0xaboomar:
This PR introduces the following major refactors:

- Create and configure a shared http.Client with connection pooling to be reused across the application.
- Encapsulate GTFS-RT data into a thread-safe RealtimeStore struct.
- Ensure realtime data is fetched once per server and passed to all dependent functions (e.g., telemetry, invalid vehicles, out-of-bounds tracking).
- Remove redundant GTFS-RT API calls and centralize fetch-and-parse logic in FetchAndStoreGTFSRTFeed.

To Do:
- The Go SDK provides its own HTTP client for `GetAgenciesWithCoverage`; explore whether it can be configured to reuse connections via a shared `http.Transport` or injected client.

### DIFF
--- a/internal/app/client.go
+++ b/internal/app/client.go
@@ -3,8 +3,55 @@ package app
 import (
 	"net"
 	"net/http"
+	"strconv"
 	"time"
+
+	"watchdog.onebusaway.org/internal/metrics"
 )
+
+// latencyTrackingRoundTripper is a custom HTTP RoundTripper that wraps another RoundTripper
+// to measure and record the latency (duration) of each outgoing HTTP request.
+//
+// Purpose:
+// - Collect Prometheus metrics on request latency (in seconds)
+// - Label the metrics by URL, HTTP method, and response status
+// - Help monitor external API performance in systems like Watchdog
+//
+// Why use this:
+// Prometheus doesn’t automatically track request latency. Wrapping the transport lets us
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
@Abo-Omar-74  Be sure to run `go fmt ./...`!

Abo-Omar-74:
I have applied Copilot's suggestions

Abo-Omar-74:
## Follow Up

### Add internal performance metrics

The changes introduce internal latency instrumentation for outgoing HTTP requests to external APIs and improve visibility into memory usage. These metrics enhance observability without adding significant system load.

#### Tracking external API Response Times
* Added `latencyTrackingRoundTripper` to wrap the HTTP transport and measure request duration.
* Introduced `http_outgoing_request_duration_seconds` Prometheus histogram, labeled by URL, method, and status code.
* Integrated the instrumented transport into the pooled HTTP client used across the app.

#### Memory Usage Tracking
Memory usage is tracked via Prometheus's default `/metrics` handler, utilizing Go’s built-in metrics, which update as the application runs, eliminating the need for a custom polling goroutine.

aaronbrethorst (approved):
nice work 👍

PR #78, opened 2025-07-30
perf: store GTFS static data in memory instead of using disk caching

0xaboomar:
## What changed?
This PR refactors the application to store and access GTFS static data in memory using a new `StaticStore`, instead of relying on disk-based caching. This change improves performance, reduces redundant I/O operations, and simplifies the logic for accessing GTFS data throughout the codebase.

#### summary

* Introduced `StaticStore`, a thread-safe in-memory store for GTFS bundles.
* Added `StaticStore` to the main application struct and initialized it in `main`.
* Updated `DownloadAndStoreGTFSBundle` to store parsed GTFS bundles directly into memory.
* Modified `DownloadGTFSBundles`, `RefreshGTFSBundles`, and `CheckBundleExpiration` to use `StaticStore`.
* Refactored `CheckAgenciesWithCoverage` and its matching function to access data from `StaticStore`.
* Removed all disk-based GTFS cache logic and file lookups.
* Updated `CollectMetricsForServer` to rely on in-memory data.
* Renamed `setupGtfsRtServer` to `setupGtfsServer` for use with both static and real-time mock servers.
* Updated all relevant test cases to use `StaticStore` instead of cache files.
* Added `readFixture` utility to support simplified test data loading.
---
## Why these changes? 

**What is currently cached?**
GTFS Bundles are downloaded every 24 hours and stored on disk. These are accessed by various metric-related functions.

**Benefits of current approach:**

* Avoids unnecessary re-downloads.
* Reduces network usage.

**What’s wrong with the current caching usage?**

* **Inefficiency in Access:** Every 30 seconds, multiple functions load and parse the same GTFS file, leading to repeated disk I/O and memory allocations.
* **Performance Cost:** For example, if 7 functions read the same file, it results in 7 I/O operations and 7 full memory copies per configured server every 30 seconds.

**How can this be improved?**

**Proposed Change:**

* Load and parse the GTFS bundle once every 24 hours.
* Share the parsed data across all internal functions via in-memory access.

**Advantages:**

* No redundant parsing or repeated allocations.
* Improves CPU and memory efficiency without changing application behavior.

**Trade-Off:**

* The GTFS bundle remains in memory for 24 hours, but this is still far cheaper than reloading and re-parsing on every usage.
* Still better than loading 7 copies repeatedly, which effectively kept it “always in memory” anyway.


Abo-Omar-74:
@aaronbrethorst I have resolved the typos Copilot referred to.

aaronbrethorst (approved):
Looks good!

PR #82, opened 2025-08-17
perf: reduce memory footprint of GTFS data and cache metrics exposition

0xaboomar:
## Performance Improvements
Here are the measured performance improvements after applying the optimizations:
- **Before optimization**: in-use memory ranged between **70–90 MB**.
- **After optimization**: in-use memory now ranges between **5–9 MB**.

During load testing with **10K total** requests and **1K concurrent** requests:
- **Time per request**: reduced from **495 ms** to **209 ms**.
- **Max in-use memory**: reduced from **100+ MB** to **16 MB**.

During lower load with **1K total** requests and **50 concurrent** requests:
- **Time per request**: reduced from **25 ms** to **8 ms**.
---
 **Before**
 
Before

 **After**
 After

**Note:**
* These benchmarks were conducted with Watchdog monitoring two OBA servers.
* Peak points are the points where I performed load testing using Apache Bench.
* The reserved memory here is the larger virtual address space that Go pre-allocates from the OS to reduce fragmentation and minimize allocation latency.
---

## What Changed
* Added lightweight `StaticData` and `RealtimeData` models containing only required GTFS fields.
* Updated `StaticStore` and `RealtimeStore` to store these minimal models instead of full bundles, reducing memory usage.
* Introduced `CachedPromHandler` that precomputes metrics at fixed intervals and serves cached responses.

  * Avoids expensive serialization for every scrape.

## Why

* Full GTFS static/realtime bundles kept in memory were unnecessarily large.
* Prometheus scrapes were triggering full metrics serialization on every request, wasting CPU and increasing latency.


aaronbrethorst (approved):
lgtm! 👍

PR #83, opened 2025-08-22
perf: add exponential backoff for retries and HTTP requests

Abo-Omar-74:
## Changes
* Implement `BackoffStore` with jitter and max backoff
  * Use it in the main metrics collection routine, storing the `NextRetryAt` and `BackoffDelay` for each server
* Add `DoWithBackoff` helper for HTTP requests (config + GTFS)
  * Use it with loading and refreshing configuration and downloading GTFS bundles
* Refactor GTFS bundle handling with backoff and concurrency
* Update `refreshConfig` to avoid ticker/backoff overlap issue

## Note 

**We should care about double backoff if upstream has retry+backoff**

* If the upstream OBA server or load balancer already applies its own retry/backoff strategy, layering another one on top can multiply delays unnecessarily.
* Example: the server retries internally (100ms, 200ms, 400ms…), and our client also does exponential backoff (1s, 2s, 4s…).
* Worst case: overall latency balloons without adding much resilience.

### DIFF
--- a/cmd/watchdog/main.go
+++ b/cmd/watchdog/main.go
@@ -26,7 +26,7 @@ func main() {
 	// This logger will be used throughout the application for logging messages.
 	// It can be configured to log to different outputs (e.g., console, file)
 	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
-
+	logger.Info("Starting OneBusAway Watchdog", "version", version)
 	// Load environment variables for configuration
 	configAuthUser := os.Getenv("CONFIG_AUTH_USER")
 	configAuthPass := os.Getenv("CONFIG_AUTH_PASS")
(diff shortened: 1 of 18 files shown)


Abo-Omar-74:
> Two changes:
> 
> 1. There is now a merge conflict, please fix it! :)
> 2. I'd like to see some explicit, direct tests of backoff_time_store.go

I have resolved the conflict and added a test for the `DoWithBackoff` function, as it is the main function I rely on for adding backoff delays when loading the config server and downloading GTFS bundles.

The `CollectMetricsForServer` method is special because it runs every 30 seconds as part of the scheduled metrics collection routine. Instead of relying directly on `DoWithBackoff`, I use a map to track each server’s current `backoffDelay` and `nextRetryAt` time. Before pinging a server, the function checks if the current time is still before its stored `nextRetryAt` time; if so, the request is skipped. When a server fails, its backoff delay is increased exponentially and the `nextRetryAt` is updated. On success, the backoff is reset.

**Why?** 

Because the `StartMetricsCollection` function opens a routine every 30 seconds for each server that calls the `CollectMetricsForServer` method, using a standard blocking backoff could lead to overlapping goroutines and duplicate requests. 

By storing backoff state per server and aligning retries with the 30-second collection cycle, we ensure that retries happen in a predictable way, short delays roll up to 30 seconds, and longer delays are rounded to the next multiple of 30. This keeps the retry logic simple, avoids concurrency issues, and still provides the benefits of exponential backoff.

aaronbrethorst (approved):
thanks! nice work

PR #88, opened 2026-03-14
fix: validate coordinates with isValidLatLon before S2 clustering

Vishmayraj:
The getClusterID function was producing S2 cluster IDs for stops with coordinates (0,0), which is typically a placeholder for improperly formatted GTFS data.

I noticed that the geo package already includes a validation check with the function “isValidLatLon,” but this is not currently used in the S2 clustering path for the aforementioned stops.

This PR includes the validation check in the 0 and 4 cases of the S2 fallback path, as well as a test to ensure that the previous incorrect behavior is now working as expected.

### DIFF
--- a/internal/geo/geo_cluster.go
+++ b/internal/geo/geo_cluster.go
@@ -74,7 +74,8 @@ func getClusterID(stop remoteGtfs.Stop) (clusterID string, clusterType string, o
 				return root.Id, "station", true
 			}
 			return "", "", false // malformed hierarchy
-		} else if stop.Latitude != nil && stop.Longitude != nil {
+		} else if stop.Latitude != nil && stop.Longitude != nil &&
+			isValidLatLon(*stop.Latitude, *stop.Longitude) {
 			return s2ClusterID(*stop.Latitude, *stop.Longitude, s2Level), "s2", true
 		}
 	case 1: // Station
@@ -88,7 +89,8 @@ func getClusterID(stop remoteGtfs.Stop) (clusterID string, clusterType string, o
 		if stop.Parent != nil && stop.Parent.Type == 0 {
 			grandparent := stop.Parent.Parent
 			if grandparent == nil {
-				if stop.Latitude != nil && stop.Longitude != nil {
+				if stop.Latitude != nil && stop.Longitude != nil &&
+					isValidLatLon(*stop.Latitude, *stop.Longitude) {
 					return s2ClusterID(*stop.Latitude, *stop.Longitude, s2Level), "s2", true
 				}
 				return "", "", false
--- a/internal/geo/geo_cluster_test.go
+++ b/internal/geo/geo_cluster_test.go
@@ -1,0 +1,24 @@
+package geo
+
+import (
+    "testing"
+
+    remoteGtfs "github.com/jamespfennell/gtfs"
+)
+
+func float64Ptr(f float64) *float64 { return &f }
+
+func TestGetClusterID_ZeroCoordinates(t *testing.T) {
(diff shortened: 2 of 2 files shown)


0xaboomar (commented):
Hi @Vishmayraj , thanks for the PR and for taking a close look at this.

The current behavior of `getClusterID()` is correct. Since `Stop.Latitude` and `Stop.Longitude` are pointers, we check whether they are set by verifying they are not `nil`.

The two lines you’re attempting to change already correctly represent the intended behavior, as clarified in the comments:

* Line 46: If a stop has no parent but has lat/lon, it is clustered by S2 cell.
* Line 64: If a parent exists but the grandparent is missing, we fall back to clustering by S2 using the stop’s lat/lon.

That said, you’ve raised a good point regarding how we treat coordinates `(0, 0)` as invalid. I’ll revisit this decision and verify whether it still makes sense. From a quick check, it doesn’t appear to be used elsewhere, but I’ll investigate further before deciding whether to keep or remove this constraint.

Thanks again for the thoughtful observation.

internal/geo/geo_cluster.go
```diff
@@ -74,7 +74,8 @@ func getClusterID(stop remoteGtfs.Stop) (clusterID string, clusterType string, o
 				return root.Id, "station", true
 			}
 			return "", "", false // malformed hierarchy
-		} else if stop.Latitude != nil && stop.Longitude != nil {
+		} else if stop.Latitude != nil && stop.Longitude != nil &&
+			isValidLatLon(*stop.Latitude, *stop.Longitude) {
```
0xaboomar, line 78: I recommend moving the guard `isValidLatLon()` and the check of lat and lon equal nil before the switch case so we reject malformed stops early.

internal/geo/geo_cluster_test.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+    "testing"
+
+    remoteGtfs "github.com/jamespfennell/gtfs"
+)
+
+func float64Ptr(f float64) *float64 { return &f }
+
+func TestGetClusterID_ZeroCoordinates(t *testing.T) {
+    t.Run("Stop with (0,0) coordinates should not S2 cluster", func(t *testing.T) {
```
0xaboomar, line 12: Update the test to make sure that malformed stops (ones with later lat | lon or lat not between (-90 , 90) or lon not between (-180 , 180)) is rejected. Skip the (0,0) checking because we will change this making `isValidLatLon()` depends on *float type to check whether lat/lon exist or not.

0xaboomar (changes requested):
Thanks for the PR, @Vishmayraj. Please take a look at the requested changes and let me know once they’re updated.

As an additional improvement, you could update the parameters of `isValidLatLon()` to use `*float64` and adjust the callers accordingly. This additional improvement isn’t a blocker for merging, so we can proceed without it and address it later.

aaronbrethorst:
### Code review

Found 1 issue:

1. The behavior change is not reflected in `getClusterID`'s doc comment, which is the file's spec for exactly these branches. It still says a no-parent stop "has lat/lon, cluster by S2 cell" and that only "no coordinates" / "coordinates are missing" is malformed — but after this change a stop that *has* coordinates of (0,0) (or out-of-range values) is now treated as malformed and dropped. The same gap exists in the `location_type = 4` block (L64-L65). This matters here because those comment lines are precisely what the reviewer cited as the intended contract; on this file a prior review already asked that clustering decisions be documented inline ("Can you add a comment to the codebase here indicating why this decision was made the way it was?", PR #71).

https://github.com/OneBusAway/watchdog/blob/aa709b6153116aef13d5552f45c4d72d71a16add/internal/geo/geo_cluster.go#L45-L48

PR #89, opened 2026-03-17
test: Add unit tests for geo utility functions

Vishmayraj:
`geo_utils.go` had no test coverage. This PR adds `geo_utils_test.go` 
covering:

- `computeBoundingBox`: empty input, nil coordinates, single stop, 
  multiple stops, mixed valid/nil coordinates
- `isValidLatLon`: zero coordinates, valid coordinates, boundary values, 
  out-of-range values
- `BoundingBoxStore`: set/get, nonexistent server, point inside/outside 
  bounding box
- `haversineDistance`: same point, known real-world distance

### DIFF
--- a/internal/geo/geo_utils_test.go
+++ b/internal/geo/geo_utils_test.go
@@ -1,0 +1,199 @@
+package geo
+
+// geo_utils_test.go contains unit tests for the utility functions
+// in geo_utils.go, including bounding box computation, coordinate
+// validation, BoundingBoxStore operations, and haversine distance.
+//
+// Author: Zala Vishmayraj
+//
+// Run tests:
+//   go test ./internal/geo/... -v (all tests)
+//
+// Tests are isolated so failure in any will stand out.
+
+import (
+	"testing"
+
+	remoteGtfs "github.com/jamespfennell/gtfs"
+)
+
+func float64Ptr(f float64) *float64 { return &f }
+
+// Helpers
+
+func makeStop(lat, lon float64) remoteGtfs.Stop {
+	return remoteGtfs.Stop{
+		Latitude:  float64Ptr(lat),
+		Longitude: float64Ptr(lon),
+	}
+}
+
+func makeStopNoCoords() remoteGtfs.Stop {
+	return remoteGtfs.Stop{}
+}
+
+// computeBoundingBox
+
+func TestComputeBoundingBox(t *testing.T) {
+	t.Run("Empty slice returns error", func(t *testing.T) {
+		_, err := computeBoundingBox([]remoteGtfs.Stop{})
+		if err == nil {
+			t.Error("Expected error for empty slice, got nil")
+		}
+	})
+
+	t.Run("All stops have nil coordinates returns error", func(t *testing.T) {
+		stops := []remoteGtfs.Stop{makeStopNoCoords(), makeStopNoCoords()}
+		_, err := computeBoundingBox(stops)
+		if err == nil {
+			t.Error("Expected error when all stops have nil coordinates, got nil")
+		}
+	})
+
+	t.Run("Single stop min equals max", func(t *testing.T) {
+		stops := []remoteGtfs.Stop{makeStop(47.60, -122.33)}
+		bbox, err := computeBoundingBox(stops)
+		if err != nil {
+			t.Fatal(err)
+		}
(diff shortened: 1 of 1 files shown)


internal/geo/geo_utils_test.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+	})
+
+	t.Run("Known distance Seattle to New York approximately correct", func(t *testing.T) {
+		// Seattle: 47.6062, -122.3321
+		// New York: 40.7128, -74.0060
+		// Known great-circle distance: ~3,867 km
+		dist := haversineDistance(47.6062, -122.3321, 40.7128, -74.0060)
+		expectedMeters := 3867000.0
+		marginMeters := 50000.0 // 50km tolerance
```
0xaboomar, line 194: The current 50km margin is a bit too loose. The theoretical maximum error for spherical approximations (like Haversine) is roughly 0.5%, which equates to ~19km for this Seattle-to-NY distance.

I suggest tightening the margin to 25km. This is wide enough to cover the 0.5% model discrepancy while being significantly more precise than 50km.

Look at the "Accuracy" note in this source for more details: https://www.movable-type.co.uk/scripts/latlong.html

internal/geo/geo_utils_test.go
```diff
@@ -0,0 +1,199 @@
...
+// in geo_utils.go, including bounding box computation, coordinate
+// validation, BoundingBoxStore operations, and haversine distance.
+//
+// Author: Zala Vishmayraj
+//
+// Run tests:
+//   go test ./internal/geo/... -v (all tests)
+//
+// Tests are isolated so failure in any will stand out.
+
```
0xaboomar, line 13: I recommend removing the header comment block (including the Author tag and test instructions). In Go, the filename and the tests themselves should be self-documenting, and Git handles the authorship tracking. This will keep the file cleaner.

internal/geo/geo_utils_test.go
```diff
@@ -0,0 +1,199 @@
...
+			makeStop(47.60, -122.33),
+			makeStopNoCoords(),
+		}
+		bbox, err := computeBoundingBox(stops)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if bbox.MinLat != 47.60 || bbox.MaxLat != 47.60 {
+			t.Errorf("Expected only valid stop coordinates, got %v", bbox)
+		}
```
0xaboomar, line 95: I suggest adding checks for MinLon and MaxLon as well.

internal/geo/geo_utils_test.go
```diff
@@ -0,0 +1,199 @@
+package geo
```
0xaboomar, line 1: Since these geographic tests are primarily mathematical and completely independent, I suggest adding t.Parallel() to the top of each test function and subtest (t.Run).

0xaboomar (changes requested):
Hi @Vishmayraj , thanks for the PR and the work you’ve put into this.

I’ve left a few comments and requested some changes that should be addressed before we proceed with merging. Please take a look and let me know once they’re updated.

aaronbrethorst:
### Code review

Found 1 issue:

1. The test file imports `github.com/jamespfennell/gtfs`, which is no longer a dependency of this module. `main` migrated the GTFS parser to `github.com/OneBusAway/go-gtfs` in #93 (commit `1dabe72`); `go.mod` now lists only `github.com/OneBusAway/go-gtfs v1.1.1` and `go.sum` has no `jamespfennell` entries at all. As written the package will not build on current `main` — and even if the old module were re-added, `[]jamespfennell/gtfs.Stop` is a distinct named type from the `[]OneBusAway/go-gtfs.Stop` that `computeBoundingBox` accepts, so `TestComputeBoundingBox` would still fail to typecheck. Worth calling out explicitly because GitHub reports this PR as mergeable/clean: the Go workflow has never run on the head commit (only the CLA check is present), so nothing here will surface the break before it lands on `main`. Rebasing on `main` and switching the import to `remoteGtfs "github.com/OneBusAway/go-gtfs"` should be all that is needed — @0xaboomar's guidance on #93 was to keep the `remoteGtfs` alias and change only the module.

https://github.com/OneBusAway/watchdog/blob/e1d26b3c63c3c6969fdc245d58b9db5438f8309e/internal/geo/geo_utils_test.go#L14-L18

PR #90, opened 2026-03-17
test: Add unit tests for middleware

Vishmayraj:
`internal/middleware` had no test coverage. This PR adds tests for:

`SecurityHeaders`: verifies all 7 security headers are set with correct 
values, and that the next handler is called correctly.

`CachedPromHandler`: verifies empty cache fallback, correct Content-Type 
when cache is populated, and concurrent read safety (verified with -race).

`SentryMiddleware` is intentionally skipped — it wraps a third-party SDK 
with no testable surface without mocking Sentry internals.

### DIFF
--- a/internal/middleware/cached_prom_handler_test.go
+++ b/internal/middleware/cached_prom_handler_test.go
@@ -1,0 +1,86 @@
+package middleware
+
+// cached_prom_handler_test.go contains unit tests for the CachedPromHandler
+// defined in cached_prom_handler.go.
+//
+// Author: Zala Vishmayraj
+//
+// Run tests:
+//   go test ./internal/middleware/... -run TestCachedPromHandler
+//   go test ./internal/middleware/... -run TestCachedPromHandler -v
+//   go test ./internal/middleware/... -race (concurrent safety)
+
+import (
+	"context"
+	"net/http/httptest"
+	"strings"
+	"sync"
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The two `CachedPromHandler` cache tests never actually populate the cache, so they pass regardless of the caching behavior. `prometheus.NewRegistry()` returns an empty registry with no collectors, so `refreshLoop` gathers zero metric families and `promhttp` encodes nothing — `c.cache` stays `len == 0` forever, and every request keeps taking the `if len(c.cache) == 0` fallback branch to the live handler. The `time.Sleep(200 * time.Millisecond)` and its "Wait for the cache to be populated" comment therefore buy nothing, and the `Content-Type` assertion cannot distinguish the two paths because the live `promhttp` handler also negotiates `text/plain; version=0.0.4`. The same applies to the "Concurrent reads do not race" subtest, which exercises only the fallback path. Registering a real collector on the registry (e.g. `registry.MustRegister(prometheus.NewCounter(...))` and incrementing it) would make the cached branch actually reachable.

https://github.com/OneBusAway/watchdog/blob/a81129f5505c91cc6374cbeb0b855f5ef0d0d789/internal/middleware/cached_prom_handler_test.go#L44-L50

PR #93, opened 2026-05-05
Migrate to oba gtfs

### LINKED ISSUE #92: Migrate GTFS Parsing Library to OneBusAway/go-gtfs
## Context
The Watchdog application currently relies on `github.com/jamespfennell/gtfs` for parsing static GTFS bundles and real-time GTFS-RT feeds. Recently, OneBusAway has taken over maintenance of its own GTFS library: [github.com/OneBusAway/go-gtfs](https://github.com/OneBusAway/go-gtfs).

To improve alignment with the OneBusAway ecosystem and ensure long-term maintainability, we want to transition the Watchdog codebase to use `OneBusAway/go-gtfs` instead.

## Objectives
1.  **Dependency Replacement**: Remove `github.com/jamespfennell/gtfs` from `go.mod` and replace it with `github.com/OneBusAway/go-gtfs`.
2.  **Refactor Parsing Logic**: Update internal services and models to match the API of the new library.
3.  **Enhance Test Assertions**: The current tests primarily check for record counts and IDs. As part of this migration, you must **expand the unit tests** by adding more detailed assertions and verifying new fields. You should decide which fields are critical to verify (e.g., coordinates, timestamps, names, or descriptive fields) for both static and real-time data to ensure that the new library is performing a complete and accurate parse.
4.  **Ensure Feature Parity**: The migration must not change any current functionality. All existing metric calculations and spatial checks must work exactly as before.

## Verification
Existing tests are built to be self-sufficient and do **not** require a live GTFS server.

### 1. Automated Tests
Run `go test ./internal/gtfs/...` and ensure the following tests pass with your **newly expanded assertions**:
*   `TestDownloadGTFSBundle`: Verifies static GTFS parsing using `testdata/gtfs.zip`.
*   `TestFetchAndStoreGTFSRTFeed`: Verifies Real-time parsing using `testdata/gtfs_rt_feed_vehicles.pb`.
*   `TestGetStopLocationsByIDs`: Verifies data retrieval from the parsed bundle.

### 2. Manual Smoke Test 
After the automated tests pass, verify the application health in the Docker environment:
1.  Run `docker-compose up --build`.
2.  Check the application logs to ensure there are no panics or GTFS-related errors.
3.  Visit the metrics endpoint (e.g., `localhost:4000/metrics`) to verify that metrics are being generated and populated as expected.

## Definition of Done
- [ ] `OneBusAway/go-gtfs` is the only GTFS library in `go.mod`.
- [ ] Unit tests are expanded with deeper data verification logic and all tests pass.
- [ ] Application starts and collects metrics successfully in the Docker environment.
- [ ] No regressions in data accuracy or metric generation.

## References
*   **Current Engine:** [jamespfennell/gtfs](https://github.com/jamespfennell/gtfs)
*   **Target Library:** [OneBusAway/go-gtfs](https://github.com/OneBusAway/go-gtfs)

MohamedKamal000:
closes #92 

for the gtfs tests i did the following:

currently watchdog monitors `stops`, `agencies` for static data and `vehiclePosition` for gtfs rt 

i covered these fields for agencies since they are required [here](https://gtfs.org/documentation/schedule/reference/#agencytxt): 
* `agency_name`
* `agency_url`
* `agency_timezone`
* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * 

for stops i covered:
* `stop_name`
* `stop_lat`
*  `stop_lon`

I compared the result based on some of the data that already exist in the gtfs test
* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * 

i wrote an assert helper function that asserts all fields a vehicle has

note: watchdog also uses `calendar.txt` file data but it uses it as a work around to get the earliest and latest service end dates, I didn't find any tests written related to it should I write unit tests for these as well ? 

@0xaboomar take a look and let me know what else might be needed/changed

## Summary by CodeRabbit

* **Chores**
  * Switched the internal GTFS library to a compatible alternative for static and realtime feed processing.

* **Tests**
  * Updated and expanded tests and test helpers to validate compatibility with the new GTFS library, adding stronger assertions for agencies, stops, vehicles, positions, timestamps, and realtime parsing.

### DIFF
--- a/internal/app/test_helpers.go
+++ b/internal/app/test_helpers.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 	"time"
 
-	obaGtfs "github.com/OneBusAway/go-gtfs"
+	remoteGtfs "github.com/OneBusAway/go-gtfs"
 	"github.com/prometheus/client_golang/prometheus"
 	"watchdog.onebusaway.org/internal/config"
 	"watchdog.onebusaway.org/internal/geo"
@@ -49,7 +49,7 @@ func newTestApplication(t *testing.T) *Application {
 	if err != nil {
 		t.Fatalf("Failed to read GTFS fixture: %v", err)
 	}
-	staticBundle, err := obaGtfs.ParseStatic(fileBytes, obaGtfs.ParseStaticOptions{})
+	staticBundle, err := remoteGtfs.ParseStatic(fileBytes, remoteGtfs.ParseStaticOptions{})
 	if err != nil {
 		t.Fatalf("Failed to parse GTFS data: %v", err)
 	}
@@ -75,7 +75,7 @@ func newTestApplication(t *testing.T) *Application {
 	if err != nil {
 		t.Fatalf("Failed to read GTFS-RT fixture: %v", err)
 	}
-	gtfsRT, err := obaGtfs.ParseRealtime(data, &obaGtfs.ParseRealtimeOptions{})
+	gtfsRT, err := remoteGtfs.ParseRealtime(data, &remoteGtfs.ParseRealtimeOptions{})
 	if err != nil {
 		t.Fatalf("Failed to parse GTFS-RT data: %v", err)
 	}
--- a/internal/app/test_helpers.go
+++ b/internal/app/test_helpers.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 	"time"
 
-	remoteGtfs "github.com/jamespfennell/gtfs"
+	obaGtfs "github.com/OneBusAway/go-gtfs"
 	"github.com/prometheus/client_golang/prometheus"
 	"watchdog.onebusaway.org/internal/config"
 	"watchdog.onebusaway.org/internal/geo"
@@ -49,7 +49,7 @@ func newTestApplication(t *testing.T) *Application {
 	if err != nil {
 		t.Fatalf("Failed to read GTFS fixture: %v", err)
 	}
-	staticBundle, err := remoteGtfs.ParseStatic(fileBytes, remoteGtfs.ParseStaticOptions{})
+	staticBundle, err := obaGtfs.ParseStatic(fileBytes, obaGtfs.ParseStaticOptions{})
 	if err != nil {
 		t.Fatalf("Failed to parse GTFS data: %v", err)
 	}
@@ -75,7 +75,7 @@ func newTestApplication(t *testing.T) *Application {
 	if err != nil {
 		t.Fatalf("Failed to read GTFS-RT fixture: %v", err)
 	}
-	gtfsRT, err := remoteGtfs.ParseRealtime(data, &remoteGtfs.ParseRealtimeOptions{})
+	gtfsRT, err := obaGtfs.ParseRealtime(data, &obaGtfs.ParseRealtimeOptions{})
 	if err != nil {
 		t.Fatalf("Failed to parse GTFS-RT data: %v", err)
 	}
--- a/internal/gtfs/gtfs_bundles_test.go
+++ b/internal/gtfs/gtfs_bundles_test.go
@@ -56,6 +56,46 @@ func TestDownloadGTFSBundle(t *testing.T) {
 		if staticBundle == nil {
 			t.Fatal("static data retrieved from the store is nil; expected non-nil value")
 		}
+		data := readFixture(t, "gtfs.zip")
+		expectedStaticData, err := remoteGtfs.ParseStatic(data, remoteGtfs.ParseStaticOptions{})
+		if err != nil {
+			t.Fatalf("failed to parse expected GTFS static data from fixture: %v", err)
+		}
+		if expectedStaticData == nil {
+			t.Fatal("parsed expected static data is nil; expected valid GTFS data")
+		}
+		if expectedStaticData.Agencies == nil {
+			t.Fatal("expected static data has nil Agencies slice; expected it to be parsed")
+		}
+
+		// For simplicity, we validate the content of agency.txt by comparing the agency IDs.
+		// We assume that if the agency IDs match, the GTFS static data was parsed and stored correctly.
+		// This level of verification is sufficient for this test.
+		//
+		// Note: We rely on agency.txt as it is a required GTFS file.
+		// Make sure the test data provided includes a non-empty agency.txt file.
+
+		if len(expectedStaticData.Agencies) != len(staticBundle.Agencies) {
+			t.Fatalf("expected %d agencies, got %d", len(expectedStaticData.Agencies), len(staticBundle.Agencies))
+		}
+		if len(expectedStaticData.Agencies) == 0 {
+			t.Fatal("expected Agencies slice is empty; can't verify content consistency")
+		}
+		expectedAgencyIDs := make(map[string]struct{})
+		for _, agency := range expectedStaticData.Agencies {
+			expectedAgencyIDs[agency.Id] = struct{}{}
+		}
+		if staticBundle.Agencies == nil {
+			t.Fatal("stored static data has nil Agencies slice; expected it to be populated")
+		}
+		if len(staticBundle.Agencies) == 0 {
+			t.Fatal("stored Agencies slice is empty; static data likely not parsed correctly")
+		}
+		for _, agency := range staticBundle.Agencies {
+			if _, ok := expectedAgencyIDs[agency.Id]; !ok {
+				t.Fatalf("unexpected agency ID %s found in stored static data", agency.Id)
+			}
+		}
 	})
 
 	t.Run("Invalid URL", func(t *testing.T) {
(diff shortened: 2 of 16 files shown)


internal/gtfs/test_helpers.go
```diff
@@ -55,3 +57,89 @@ func readFixture(t *testing.T, fixturePath string) []byte {
...
+	t.Helper()
+	if expected == nil {
+		t.Fatal("expected vehicle must not be nil")
+	}
+	if actual == nil {
+		t.Errorf("actual vehicle is nil")
+		return
+
+	}
+	if expected.ID != nil {
```
0xaboomar: Note that the case where expected.ID == nil while actual.ID != nil is not handled, and similarly for StopID, CurrentStopSequence, Timestamp, and Position.

internal/gtfs/test_helpers.go
```diff
@@ -55,3 +57,89 @@ func readFixture(t *testing.T, fixturePath string) []byte {
...
+		if actual.ID == nil {
+			t.Errorf("vehicle ID missing")
+		} else if expected.ID.ID != actual.ID.ID {
+			t.Errorf("vehicle ID mismatch: expected %s, got %s",
+				expected.ID.ID, actual.ID.ID)
+		}
+	}
+
```
0xaboomar, line 102: The current structure is quite nested , which makes it harder to scan and maintain. Consider flattening the conditions and/or extracting small helper functions for pointer comparisons to improve readability and reduce duplication.

0xaboomar (changes requested):
Thanks for the PR @MohamedKamal000, overall it looks good and well-structured. I left a few comments mainly around handling missing vs unexpected fields and some readability improvements to make the assertions easier to follow and maintain.

internal/app/test_helpers.go
```diff
@@ -7,7 +7,7 @@ import (
 	"testing"
 	"time"
 
-	remoteGtfs "github.com/jamespfennell/gtfs"
+	obaGtfs "github.com/OneBusAway/go-gtfs"
```
0xaboomar: I think we can keep the name `remoteGtfs` and only change the module being used. The reason is that in Watchdog we already have an internal module called `gtfs`, so `remoteGtfs` helps distinguish it from the remote one.

Moreover, some files currently import only the external GTFS module, so we didn’t give it an alternative name. For clarity and consistency, I think it’s better to update those imports as well.

0xaboomar:
> note: watchdog also uses calendar.txt file data but it uses it as a work around to get the earliest and latest service end dates, I didn't find any tests written related to it should I write unit tests for these as well ?

For sure @MohamedKamal000 , It would be a great!

internal/gtfs/gtfs_bundles_test.go
```diff
@@ -82,9 +82,21 @@ func TestDownloadGTFSBundle(t *testing.T) {
 		if len(expectedStaticData.Agencies) == 0 {
 			t.Fatal("expected Agencies slice is empty; can't verify content consistency")
 		}
-		expectedAgencyIDs := make(map[string]struct{})
+		expectedAgencyIDs := make(map[string]struct {
```
0xaboomar: It’d be better to move these checks from `TestDownloadGTFSBundle` into a separate parsing logic test.

internal/gtfs/gtfs_bundles_test.go
```diff
@@ -113,23 +139,57 @@ func TestGetStopLocationsByIDs(t *testing.T) {
...
 		t.Fatal("failed to parse gtfs static data")
 	}
 	staticData := models.NewStaticData(staticBundle)
 	staticStore := NewStaticStore()
 	staticStore.Set(server.ID, staticData)
 
-	t.Run("Valid stop IDs", func(t *testing.T) {
+	t.Run("Valid stops", func(t *testing.T) {
 		stopIDs := []string{"11060", "1108"} // Make sure these exist in your test GTFS
+		stopsData := map[string]struct {
```
0xaboomar: It’d be better to move these checks from `TestGetStopLocationsByIDs` into a separate parsing logic test.

0xaboomar (changes requested):
I think instead of expanding `TestDownloadGTFSBundle` and `TestGetStopLocationsByIDs`whose main purpose is to test the behavior of `DownloadGTFSBundle` and `GetStopLocationsByIDs`; It would be better to keep the tests modular and add a separate test for the parsing logic.

Take a look at [internal/models/gtfs_models.go](https://github.com/OneBusAway/watchdog/blob/main/internal/models/gtfs_models.go)
You’ll see that we currently rely on `Stops`, `Agencies`, and `Services` for static GTFS, and `Vehicle` for realtime GTFS.

The parsing tests should cover all of these, including their attributes. For `Vehicle`, there are fields with complex types like `Trip` and `Position`.
We don’t need to fully test all nested structures, just ensure the nesting behavior works. 

So we can skip the more complex ones and include only the `Position` field (since it’s simpler and all its attributes are primitive types), along with the other primitive fields and type alias. That should be sufficient for this approach.

MohamedKamal000:
hey @0xaboomar  ty for the review
I have made all your requested changes take a look at them now and let me know

for the agency parsing, I hard coded the values that exist in the test data because I think making a `mockServer` then downloading the bundle (which uses `parseStatic` internally) and comparing it to the `praseStatic` result is not reasonable

MohamedKamal000:
> Take a look at [internal/models/gtfs_models.go](https://github.com/OneBusAway/watchdog/blob/main/internal/models/gtfs_models.go) You’ll see that we currently rely on `Stops`, `Agencies`, and `Services` for static GTFS, and `Vehicle` for realtime GTFS.

ya I noticed this as written in this comment 
```
// IMPORTANT:
// In the future, we may need to extend this structure
// to include more fields from the GTFS Static bundle.
// Don't forget to include them here
```

if you think it is time to expand these to monitor more gtfs data you can raise some github issues to track their progress or something and i would be happy to help you in the development in my free time

0xaboomar:
> for the agency parsing, I hard coded the values that exist in the test data because I think making a mockServer then downloading the bundle (which uses parseStatic internally) and comparing it to the praseStatic result is not reasonable

Right move for parsing test @MohamedKamal000. `TestDownloadGTFSBundle` was testing the behavior of the download function, not the parser. In this test, we trust the parser since the goal is to verify whether the function downloaded the same data or not. That’s why we were parsing both the expected and actual data using the same logic.

internal/gtfs/test_helpers.go
```diff
@@ -55,3 +58,82 @@ func readFixture(t *testing.T, fixturePath string) []byte {
 
 	return data
 }
+
+func assertPtr[T comparable](t *testing.T, expected *T, actual *T, field string, equal func(a, b T) bool) {
```
0xaboomar, line 62: This is smart, great work!

internal/gtfs/gtfs_bundles_test.go
```diff
@@ -56,57 +56,167 @@ func TestDownloadGTFSBundle(t *testing.T) {
...
+	if len(expectedStaticAgencies) != len(staticBundle.Agencies) {
+		t.Fatalf("expected %d agencies, got %d", len(expectedStaticAgencies), len(staticBundle.Agencies))
+	}
+
+	expectedAgencyIDs := make(map[string]struct {
+		Name      string
+		TimeZone  string
+		AgencyUrl string
+	})
+
```
0xaboomar, line 143: You can make the type of `expectedAgencyIDs` map a string to `remoteGtfs.Agency` and include all agency attributes. I know the ID will be redundant in both the key and value, but that’s okay. This way, we can ensure that all fields are parsed correctly.

internal/gtfs/gtfs_bundles_test.go
```diff
@@ -56,57 +56,167 @@ func TestDownloadGTFSBundle(t *testing.T) {
...
+		expectedAgencyIDs[agency.Id] = struct {
+			Name      string
+			TimeZone  string
+			AgencyUrl string
+		}{
+			Name:      agency.Name,
+			TimeZone:  agency.Timezone,
+			AgencyUrl: agency.Url,
 		}
-		if expectedStaticData.Agencies == nil {
```
0xaboomar, line 68: I think we should revert the removed part, as the goal of `TestDownloadGTFSBundles` is to check that we are downloading the GTFS bundle correctly, not testing the parsing logic itself. We simply trust the parser and use it to parse both the actual and expected data.

Since we trust the parsing logic, checking that one of the files matches is enough to ensure we downloaded the expected data. That’s why we were only checking `agency.id` from the agency file for simplicity.

The parsing logic should instead have more comprehensive tests, which is exactly the work you are doing now. Thanks for that.

internal/gtfs/test_helpers.go
```diff
@@ -55,3 +58,82 @@ func readFixture(t *testing.T, fixturePath string) []byte {
...
+	}
+	if !equal(*expected, *actual) {
+		t.Errorf("%s mismatch: expected %v, got %v", field, *expected, *actual)
+	}
+}
+
+func assertVehicle(t *testing.T, actual *remoteGtfs.Vehicle, expected *remoteGtfs.Vehicle) {
+	t.Helper()
+	if expected == nil {
+		t.Fatal("expected vehicle must not be nil")
```
0xaboomar, line 84: The one thing that remains is to make the error messages clearer and include both the expected and actual values in every message, as that will make debugging much easier. This applies to all the error messages you added.

0xaboomar (changes requested):
Hey @MohamedKamal000 , thanks for the effort you are putting into this. It’s looking pretty good overall. I’ve added some changes that need to be addressed before we proceed with merging.

0xaboomar:
> ya I noticed this as written in this comment
> 
> // IMPORTANT:
> // In the future, we may need to extend this structure
> // to include more fields from the GTFS Static bundle.
> // Don't forget to include them here
> if you think it is time to expand these to monitor more gtfs data you can raise some github issues to track their progress or something and i would be happy to help you in the development in my free time

For sure, I’ll think more about how we can expand our metrics in a way that is actually useful and probably open some issues soon. I’d definitely like seeing you contributing to them as well.

MohamedKamal000:
reverted the `TestDownloadGTFSBundle` function as you mentioned and applied other suggestions

MohamedKamal000:
also I'm not sure but don't we need a github workflow for running tests on pr push ? if you think we need one i might try to raise a pr for this because i want to learn more about github workflows and try do one myself, your review would also be valuable and could help me.

MohamedKamal000:
> also I'm not sure but don't we need a github workflow for running tests on pr push ? if you think we need one i might try to raise a pr for this because i want to learn more about github workflows and try do one myself, your review would also be valuable and could help me.

oh nvm i think we already have one but it is enabled on pushes to main only i think, didn't catch that

0xaboomar:
LGTM!

PR #129, opened 2026-08-04
feat: prometheus alerting rules

0xaboomar:
## Summary by CodeRabbit

* **New Features**
  * Added centralized alert management with grouped notifications, delayed delivery, repeat reminders, and resolved-alert updates.
  * Added monitoring alerts for stale feeds, low trip- and stop-matching quality, and unavailable watchdog services.
  * Configured monitoring to automatically discover alert rules and forward notifications.

* **Operations**
  * Added the alert management service to the monitoring environment for improved incident visibility.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,8 +19,21 @@ services:
     container_name: prometheus
     volumes:
       - ./prometheus/prometheus.yml:/etc/prometheus/prometheus.yml
+      - ./prometheus/rules:/etc/prometheus/rules
     ports:
       - "9090:9090"
+    networks:
+      - observability
+    depends_on:
+      - alertmanager
+
+  alertmanager:
+    image: prom/alertmanager:latest
+    container_name: alertmanager
+    volumes:
+      - ./prometheus/alertmanager.yml:/etc/alertmanager/alertmanager.yml
+    ports:
+      - "9093:9093"
     networks:
       - observability
 
--- a/prometheus/alertmanager.yml
+++ b/prometheus/alertmanager.yml
@@ -1,0 +1,12 @@
+route:
+  receiver: "webhook"
+  group_by: ["alertname", "severity"]
+  group_wait: 30s
+  group_interval: 5m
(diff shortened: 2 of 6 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is clean. I checked every metric name and label in the rules against `internal/metrics/metrics.go` on `main` — `oba_time_since_last_update_seconds`, `oba_realtime_trip_match_ratio`, and `oba_stop_match_ratio` all exist and all carry the `server` label your annotations interpolate, and `up{job="watchdog"}` matches the existing scrape job name. The rule glob, the bind mount, and the `alertmanager:9093` target line up, and the whole thing validates under `promtool check rules`, `promtool check config`, `amtool check-config`, and `docker compose config`. The `absent()` arm on `WatchdogDown` plus `keep_firing_for` is a nice detail.

Merging as-is. A few follow-ups worth picking up separately, none of which should hold this:

- The webhook receiver points at `http://host.internal/alerts`, which nothing in the repo serves and which won't resolve on plain Linux Docker Engine. Worth a comment in `alertmanager.yml` marking it as a placeholder.
- All three OBA metrics are `(server, agency)`-labeled but the annotations only interpolate `server`, so a multi-agency server produces alerts with byte-identical summaries — and `group_by: ["alertname", "severity"]` groups them into one notification. Adding `{{ $labels.agency }}` fixes both.
- Alertmanager has no data volume, so silences and the notification log are lost on container recreate.
- `FeedStale` can't catch a fully unreachable OBA server: `oba_time_since_last_update_seconds` is only set when the poll succeeds, so the gauge freezes at its last good value rather than climbing past 900. An `oba_api_status == 0` rule would close that gap.
- Heads up that #131 removes the `server` label from all three of these metrics in favor of `agency_id`. These rules are correct today; they'll need updating when that lands.

0xaboomar:
Thanks for merging, @aaronbrethorst! And no worries, I’ll take care of the points raised.

PR #130, opened 2026-08-06
fix: report invalid servers once

0xaboomar:
## Summary by CodeRabbit

- **New Features**
  - Invalid or duplicate server entries are tracked consistently across startup and configuration refreshes.
  - Server recovery and recurring configuration issues are reported more consistently.
  - Configuration validation provides clearer handling for feed and agency details.
  - The application now shuts down gracefully when interrupted.

- **Bug Fixes**
  - Invalid entries are re-evaluated during later refreshes, allowing recovered servers to return automatically.

- **Documentation**
  - Updated configuration examples to remove the no-longer-supported `agency_ids` field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,14 +34,13 @@ Watchdog requires a configuration file (`config.json`) before running. Even plac
       "trip_update_url": "https://trip1.example.com",
       "vehicle_position_url": "https://vehicle1.example.com",
       "gtfs_rt_api_key": "api-key-1",
-      "gtfs_rt_api_value": "api-value-1",
-      "agency_ids": []
+      "gtfs_rt_api_value": "api-value-1"
     }]
   }
 ]
 ```
 
-The top-level `agency_id` decides what each entry observes: set it to track that agency. OBA API probes query only that agency, though static and realtime feed data is not filtered by agency; leave it out to track the server and every agency it serves. `agency_ids` on a feed is optional and unused by Watchdog today. See [Two observation modes](#two-observation-modes-agency-vs-server) below for both config shapes.
+The top-level `agency_id` decides what each entry observes: set it to track that agency. OBA API probes query only that agency, though static and realtime feed data is not filtered by agency; leave it out to track the server and every agency it serves. See [Two observation modes](#two-observation-modes-agency-vs-server) below for both config shapes.
 
 #### Ways to Provide the Config File
 
@@ -98,8 +97,7 @@ There are two ways Watchdog can observe an OBA deployment. The choice is made **
     "trip_update_url": "https://trip1.example.com",
     "vehicle_position_url": "https://vehicle1.example.com",
     "gtfs_rt_api_key": "api-key-1",
-    "gtfs_rt_api_value": "api-value-1",
-    "agency_ids": ["agency-1"]
+    "gtfs_rt_api_value": "api-value-1"
   }]
 }
 ```
@@ -116,8 +114,7 @@ There are two ways Watchdog can observe an OBA deployment. The choice is made **
     "trip_update_url": "https://trip2.example.com",
     "vehicle_position_url": "https://vehicle2.example.com",
     "gtfs_rt_api_key": "api-key-2",
-    "gtfs_rt_api_value": "api-value-2",
-    "agency_ids": []
+    "gtfs_rt_api_value": "api-value-2"
   }]
 }
 ```
--- a/cmd/watchdog/main.go
+++ b/cmd/watchdog/main.go
@@ -88,11 +88,14 @@ func main() {
 	// Load the configuration from the specified source
 	// If a config file is specified, load it from disk.
 	// If a config URL is specified, fetch it over HTTP(S).
+	// The dropped-servers store is shared between the initial load and the
+	// periodic refresh so an invalid server is reported to Sentry only once.
+	droppedStore := config.NewDroppedServersStore()
 	var servers []models.ObaServer
 	if *configFile != "" {
-		servers, err = config.LoadConfigFromFile(*configFile)
+		servers, err = config.LoadConfigFromFile(*configFile, droppedStore)
 	} else if *configURL != "" {
-		servers, err = config.LoadConfigFromURL(ctx, client, *configURL, configAuthUser, configAuthPass, 20)
+		servers, err = config.LoadConfigFromURL(ctx, client, *configURL, configAuthUser, configAuthPass, 20, droppedStore)
 	}
 
 	if err != nil {
@@ -115,7 +118,7 @@ func main() {
 	// and the required dependencies.
 	// this New() function is critical in understanding how we structure the application take a look at it.
 	// and also take a look at service file in each package to see the dependencies and the exposed methods and function.
-	app := app.New(&cfg, logger, client, version)
+	app := app.New(&cfg, logger, client, version, droppedStore)
 
 	// From here we set up all dependencies and we are ready to start business logic.
 
--- a/cmd/watchdog/main.go
+++ b/cmd/watchdog/main.go
@@ -112,8 +112,6 @@ func main() {
 	// The dropped-servers store is shared between the initial load and the
 	// periodic refresh so an invalid server is reported to Sentry only once.
 	droppedStore := config.NewDroppedServersStore()
-	// Share transition state between startup and remote refreshes.
-	droppedStore := config.NewDroppedServersStore()
 	var servers []models.ObaServer
 	if *configFile != "" {
 		servers, err = config.LoadConfigFromFile(*configFile, logger, droppedStore)
--- a/cmd/watchdog/main.go
+++ b/cmd/watchdog/main.go
@@ -109,11 +109,16 @@ func main() {
 	// Load the configuration from the specified source
 	// If a config file is specified, load it from disk.
 	// If a config URL is specified, fetch it over HTTP(S).
+	// The dropped-servers store is shared between the initial load and the
+	// periodic refresh so an invalid server is reported to Sentry only once.
+	droppedStore := config.NewDroppedServersStore()
+	// Share transition state between startup and remote refreshes.
+	droppedStore := config.NewDroppedServersStore()
 	var servers []models.ObaServer
 	if *configFile != "" {
-		servers, err = config.LoadConfigFromFile(*configFile, logger)
+		servers, err = config.LoadConfigFromFile(*configFile, logger, droppedStore)
 	} else if *configURL != "" {
-		servers, err = config.LoadConfigFromURL(ctx, client, *configURL, configAuthUser, configAuthPass, 20, logger)
+		servers, err = config.LoadConfigFromURL(ctx, client, *configURL, configAuthUser, configAuthPass, 20, logger, droppedStore)
 	}
 
 	if err != nil {
@@ -136,7 +141,7 @@ func main() {
 	// and the required dependencies.
 	// this New() function is critical in understanding how we structure the application take a look at it.
 	// and also take a look at service file in each package to see the dependencies and the exposed methods and function.
-	app := app.New(&cfg, logger, client, version)
+	app := app.New(&cfg, logger, client, version, droppedStore)
 
 	// Report the agencies Watchdog is tracking (those that passed config
 	// validation). This runs once at startup; it is re-triggered only when the
--- a/config.json.template
+++ b/config.json.template
@@ -10,8 +10,7 @@
       "trip_update_url": "https://trip1.example.com",
       "vehicle_position_url": "https://vehicle1.example.com",
       "gtfs_rt_api_key": "api-key-1",
-      "gtfs_rt_api_value": "api-value-1",
-      "agency_ids": ["agency-1"]
+      "gtfs_rt_api_value": "api-value-1"
     }]
   },
   {
@@ -23,8 +22,7 @@
       "trip_update_url": "https://trip2.example.com",
       "vehicle_position_url": "https://vehicle2.example.com",
       "gtfs_rt_api_key": "api-key-2",
-      "gtfs_rt_api_value": "api-value-2",
-      "agency_ids": []
+      "gtfs_rt_api_value": "api-value-2"
     }]
   }
 ]
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -25,7 +25,7 @@ type Application struct {
 
 // New creates and wires all dependencies for the Application.
 // Accepts config, logger, client, and version as arguments.
-func New(cfg *config.Config, logger *slog.Logger, client *http.Client, version string) *Application {
+func New(cfg *config.Config, logger *slog.Logger, client *http.Client, version string, droppedStore *config.DroppedServersStore) *Application {
 
 	staticStore := gtfs.NewStaticStore()
 	realtimeStore := gtfs.NewRealtimeStore()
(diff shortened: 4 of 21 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `rejectDuplicateServerIDs` never clears `reportedDuplicates` when a duplicate is *resolved* while the ID stays in the config, so a duplicate that is fixed and later reintroduced is silently swallowed forever. The prune loop keys on `present`, which is populated for **every** server ID (line 111), not just the ones that were duplicated this cycle. Trace: `[A(id=1), A'(id=1)]` reports and sets `reportedDuplicates={1}`; operator fixes it to `[A(id=1)]`, but `present={1}` so `1` is not pruned; a later regression back to `[A(id=1), A'(id=1)]` reports nothing. The doc comment states the intent ("a duplicate that reappears later reports again") and `TestRejectDuplicateServerIDs/prunes duplicate IDs that leave the config so they report again` asserts it, but that test only exercises the case where the ID disappears from the config entirely (`validServer()` has ID 1, `dup` has ID 7) — not the normal fix-then-regress path. `Reconcile` handles the analogous transition correctly by deleting from `reported` on recovery; this path is missing the equivalent. Fix: track the IDs that were actually duplicated this cycle and prune `reportedDuplicates` against that set instead of `present`.

https://github.com/OneBusAway/watchdog/blob/45ea07574b37e1c4ee1ad4e4d048bab5f6d9872a/internal/config/dropped_servers_store.go#L110-L137

aaronbrethorst (changes requested):
Thanks for this — the design is right, and the implementation quality is high. Edge-triggering the Sentry reports off the valid→invalid transition is exactly the fix this needed, the info-level recovery report is a nice touch, and the store is wired through `app.New()` → `ConfigService` the way the rest of the codebase does it. The `rejectDuplicateServerIDs` pre-pass is a subtle thing to have anticipated, and the doc comments explaining *why* it must run before `Reconcile` are the kind of comment I wish more code had.

Two things before this can land.

## 1. `reportedDuplicates` is never cleared when a duplicate is fixed but the ID stays

In `rejectDuplicateServerIDs`, `present` is populated for every server ID:

```go
for _, server := range servers {
    present[server.ID] = struct{}{}   // every ID, duplicated or not
```

but the prune loop at the bottom only deletes from `reportedDuplicates` when the ID is absent from `present`. So the entry clears only when the ID disappears from the config *entirely*:

- Cycle 1: `[A(id=1), A'(id=1)]` → duplicate reported, `reportedDuplicates = {1}`
- Cycle 2: someone fixes it, config is `[A(id=1)]` → `present = {1}`, so the entry survives
- Cycle 3: the duplicate is reintroduced → already in `reportedDuplicates` → silent, permanently

That's the failure mode this PR exists to prevent, just moved up a level. Your doc comment says "a duplicate that reappears later reports again," which is the behavior I want — it's just not what the code does today.

The fix is small: track the IDs actually duplicated this cycle and prune against that set rather than `present`.

The existing test passes because `dup` uses ID 7 while `validServer()` is ID 1, so the ID does fully disappear. Worth adding a case where the ID stays present.

## 2. CI has never run on this branch

There is no workflow run at all for `fix/report-invalid-servers-once`. Actions is working repo-wide — #129 and #131 both ran — it just never triggered here. Push an empty commit or re-push to kick it off; I'd like to see tests green before merging.

## Smaller things, your call, not blockers

- `rejectDuplicateServerIDs` runs before `ValidateServer`, so two entries with a missing or null `id` both land on `ID == 0`. The second is reported as "duplicate server id 0", which misattributes a missing-field problem as an ID collision, and a third id-less entry is dropped with no report at all. Previously each was reported individually.
- `reported` is keyed by ID alone, so a server that becomes invalid for a *new* reason (was missing `gtfs_url`, now missing `agency_id`) stays silent. That may be the semantics you intended — I just want it to be a deliberate choice.

One sequencing note that isn't your fault: this and #131 can't both merge as-is. #131 deletes `ObaServer.ID`, which this store is keyed on, and both rewrite `filterValidServers`. I'd like #131 to go first since it's the larger structural change, which would mean rebasing this onto the `agency_id` key.

Happy to re-review as soon as the prune fix and a green CI run are up.

aaronbrethorst (changes requested):
The two blockers are genuinely fixed, and I verified them rather than reading the commit messages.

- The prune now works. `duplicated` is built from the per-cycle `counts[identity] > 1` pre-pass and the loop at the bottom deletes from `reportedDuplicates` when the identity is no longer duplicated this cycle, not when it disappears from `present`. The `two → one → two` cycle reports twice, which is what the doc comment promised all along.
- `TestReconcileReportsDuplicateAgainAfterItIsFixed` is a real regression test: the shared identity is present in all three cycles, so it fails under the old `present`-based prune instead of passing trivially.
- CI is green on the rebased head, and the #131 sequencing problem resolved itself when #131 and #133 landed. Keying on `models.ServerKey` instead of the deleted `ObaServer.ID` is the right landing spot, and routing id-less entries through a raw identity with `duplicateEligible == false` cleanly fixes the "duplicate server id 0" misattribution I raised as a non-blocker.

One new thing I'd like fixed before this lands, because it moves in the wrong direction for a monitoring service.

## A valid server is now dropped when a malformed entry shares its identity

The duplicate check runs before `decodeServerEntry`, and `seen[identity]` is marked on the way past regardless of whether the entry decodes. So for a config containing `[E1 malformed, E2 valid]` that share an `(oba_base_url, agency_id)`:

- E1 marks `seen[X]`, then fails to decode and is reported as invalid.
- E2 hits the `seen[X]` branch, is reported as a duplicate, and is dropped without ever being decoded.

Both copies are gone and the server stops being monitored. On `main` this doesn't happen: `decodeServers` runs `decodeServerEntry` first and `continue`s on failure before `seenServers` is ever touched, so the malformed copy never claims the slot and the valid one survives.

That's a silent reduction in what Watchdog watches, which is the same failure class as the frozen-series problem `PruneStaleServers` exists to prevent — the fleet looks fine because nothing is reporting on it. It needs a duplicated identity whose first copy is malformed, so it isn't common, but handling malformed entries gracefully is the entire premise of this PR.

The fix shouldn't cost you the raw-identity mechanism you added, which I do want to keep. Keep the `counts` pre-pass for the reporting decision, but only let an entry claim `seen[identity]` after `decodeServerEntry` succeeds, so a malformed entry can't reserve an identity on behalf of a valid one.

Smaller things, none blocking:

- `serverTagsFromRaw` in `internal/config/compat.go` is dead now — this PR deleted its only caller. Neither the compiler nor `go vet` will tell you, so it'll sit there.
- The recovery report hardcodes only `agency_id`/`agency_name`, while the invalid and duplicate reports use the `tags` map that includes `server_name`. The error event is correlatable per server and the matching recovery event isn't, which is awkward when you're trying to pair them.
- The duplicate `logger.Error` sits inside the `!alreadyReported` guard, so on cycles 2..N a duplicate is dropped with no local output at all. Once-per-process is a defensible reading of this PR's thesis, but it's the opposite of what I asked for on #131, so I'd rather it be deliberate than incidental.
- `Reconcile`'s doc comment covers the `reported` lifecycle but says nothing about `reportedDuplicates` or the pruning rule, which is the thing this round of review was actually about. Worth a sentence.

Also, on the "same identity, new failure reason stays silent" note from last time: unchanged, which is fine — I only wanted it to be a choice. A line in the doc comment saying so would close it out.

Separately, the CLA check is blocking this and your four other open PRs. One of two committers has signed; the unsigned commits are authored as `Mohamed Ahmed Aboomar <aboomar@Mohameds-MacBook-Air.local>`, a local machine hostname rather than a real address, so the bot's advice to add the email to your GitHub account won't work. Rewrite the authorship and force-push:

```
git push --force-with-lease
```

Fix the drop-both case and I'll re-review promptly.

aaronbrethorst:
### Code review

Found 2 issues:

1. The `[malformed, valid]` same-identity fix trades a dropped server for a permanent report loop. `s.reported[identity]` is written on the decode-failure path and read back by the recovery check *within the same pass*, so for a config containing two entries sharing an `(oba_base_url, agency_id)` where one is malformed, every refresh emits an error report for the malformed copy followed by an info "recovered" report for the valid copy — then clears the state, so the next cycle does it again. I ran this against the head SHA: 5 identical `Reconcile` calls produced 10 events, alternating `missing required fields` / `recovered: previously invalid configuration`, forever. `TestReconcileMalformedFirstValidSecondSameIdentity` asserts the single-cycle half of this as correct behavior, so it locks the flapping in. The doc comment two screens up says "invalid server already reported -> silent", which is the behavior this PR exists to deliver.

https://github.com/OneBusAway/watchdog/blob/fa7e5271f3a8bf03ec2a749e20069f42d5bcf031/internal/config/dropped_servers_store.go#L109-L121

2. `reportedDuplicates` does not reset when one copy of a duplicated identity becomes invalid, so a restored duplicate is dropped silently. The `counts` pre-pass is built from the raw entries before `decodeServerEntry` runs, so a malformed copy still contributes to `counts[identity] > 1`; the surviving valid copy therefore sets `duplicated[identity]` and the prune loop at the bottom keeps the `reportedDuplicates` entry alive. Sequence: cycle 1 `[A, A']` both valid → duplicate reported; cycle 2 `A'` goes malformed → invalid reported; cycle 3 `A'` is fixed → a server is again dropped as a duplicate with **no** Sentry report (verified at head: 3 events total, none of them the second duplicate). The doc comment added this round states the opposite — "when all but one copy are removed *or become invalid*, the entry is pruned".

https://github.com/OneBusAway/watchdog/blob/fa7e5271f3a8bf03ec2a749e20069f42d5bcf031/internal/config/dropped_servers_store.go#L56-L64

aaronbrethorst (changes requested):
The blocker from last round is genuinely fixed, and I verified it in the code rather than reading the commit message. `decodeServerEntry` now runs before the duplicate check and `seen[identity]` is only claimed after a successful decode, with a comment saying exactly why. For `[malformed, valid]` sharing an identity the valid entry survives and keeps being monitored, which is what I was worried about losing. The other four items are done too: `serverTagsFromRaw` is gone, the recovery report passes the full `tags` map so `server_name` rides along, the duplicate `logger.Error` moved outside the `!alreadyReported` guard, and the CLA is signed.

The fix works, but it opened a new hole in the same configuration, and it's the failure mode this PR exists to close. Two things, one root cause.

## 1. `[malformed, valid]` sharing an identity now flaps error/recovery forever

`s.reported[identity]` is written on the decode-failure path and read back by the recovery check **in the same pass**, which then deletes it. Walking `[E1 malformed, E2 valid]`, both with identity `X`:

- **E1** fails to decode. `reported[X]` is absent, so it's set and an **error** report fires. `continue`.
- **E2** decodes fine, reaches the recovery check, finds `reported[X]` — the entry E1 just set — deletes it and fires a **recovery** report at info level.
- End of cycle: `reported` is empty again.

Next refresh does exactly the same thing. So this config emits one error plus one false "recovered: previously invalid configuration" on **every** cycle, indefinitely. That's noisier than `main`, which at least only emitted the error, and the recovery event is a lie — nothing recovered, the malformed entry is still malformed.

`TestReconcileMalformedFirstValidSecondSameIdentity` cements it rather than catching it: it runs a single cycle, asserts exactly 2 events, and its comment describes the recovery report as correct behavior. Running it twice is what exposes the problem.

The fix is small. Snapshot the reported set at the top of `Reconcile` and drive the recovery check off the snapshot, so an identity first reported *this* cycle can't also be "recovered" in the same pass:

```go
previouslyReported := make(map[string]struct{}, len(s.reported))
for k := range s.reported {
    previouslyReported[k] = struct{}{}
}
```

then test `previouslyReported` instead of `s.reported` at the recovery branch. A genuine repair — invalid last cycle, valid this cycle — still reports exactly once.

## 2. `reportedDuplicates` isn't pruned when one copy goes invalid, so a later real duplicate is dropped silently

`counts` is built from the **raw** entries before any decode, so a malformed copy still pushes `counts[identity]` above 1. The surviving valid copy then reaches `if duplicateEligible && counts[identity] > 1` and sets `duplicated[identity]`, which keeps the bottom prune loop from clearing `reportedDuplicates[identity]`. Across three cycles:

1. `[A, A']` both valid → duplicate detected, reported, `reportedDuplicates[X]` set.
2. `[A, A']` with `A'` now malformed → `A'` is reported invalid; `A` still marks `duplicated[X]`, so `reportedDuplicates[X]` survives the prune.
3. `A'` is fixed, `[A, A']` both valid again → `A'` is dropped as a duplicate, but `reportedDuplicates[X]` already exists, so **no report fires**.

A server is actively dropped and nothing says so. That's the same shape as the bug I raised in the first round — a duplicate that reappears going silent — reached through a different door, and the doc comment you added this round promises the opposite ("or become invalid, the entry is pruned").

The cleanest fix is to set `duplicated[identity]` where the duplicate is actually detected — inside the `if _, exists := seen[identity]; exists` branch — rather than from the raw `counts` pre-pass. Then it means "two decoded entries really did claim this identity this cycle," the prune clears it when that stops being true, and a reappearing duplicate reports again.

Both issues come from the same place: malformed entries participate in `present` and `counts` but not in `seen`. Fixing the two spots above should close them together.

## Smaller, not blocking

- `TestReconcileRejectsDuplicatesBeforeValidation` now asserts that duplicates are *not* detected for malformed entries, which is the opposite of what its name says. Worth renaming while you're in there.
- The "same identity, new failure reason stays silent" note from last round still isn't mentioned in the `Reconcile` doc comment. I'd still like a line marking it as deliberate.

I checked for unbounded growth specifically since that's the risk with this kind of dedup state, and it's fine — both maps are pruned every cycle against per-cycle sets, so they stay bounded by config size.

Add a test that calls `Reconcile` twice with an unchanged `[malformed, valid]` config and asserts 1 error and 0 recoveries on the second pass, and the first issue can't come back. Happy to re-review as soon as these are up.

aaronbrethorst:
One logistical note on top of the review above: #142, #136 and #144 landed on `main` while I was going through this queue, and this branch now has merge conflicts. Please merge `main` in as part of the same pass rather than rebasing twice.

Heads up that #144 in particular touches code near yours — it threads `ctx` through the config and metrics seams, including `loadConfigFromURL`, which is one of `Reconcile`'s callers.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both blockers from the last round are genuinely fixed, and I verified them by running the sequences rather than reading the commit messages.

**The flapping error/recovery loop is gone.** `previouslyReported` is snapshotted at the top of `Reconcile` and the recovery check is driven off the snapshot, with the deferred recovery loop skipping any identity in `invalidThisCycle`. I ran five identical `Reconcile` calls over a `[malformed, valid]` config sharing an identity: **1 event total**, an error on the first cycle and silence thereafter. That same config produced 10 alternating error/recovery events last round.

**The duplicate prune is fixed at the root.** Dropping the raw `counts` pre-pass entirely and setting `duplicated[identity]` inside the `seen[identity]` branch is the better of the two fixes available — it makes the flag mean "two decoded entries really did claim this identity this cycle," which is what the prune needs. I walked the three-cycle sequence: duplicate reported, one copy goes malformed and the entry is pruned, copy is fixed and the duplicate **reports again**. That was the failure mode this PR exists to prevent, and it's closed from both directions now.

The regression tests are real, not decorative. `TestReconcileMalformedFirstValidSecondSameIdentity` runs two cycles and asserts one event; `TestReconcileReportsDuplicateAgainAfterOtherCopyBecomesInvalid` drives the full three-cycle sequence and asserts two duplicate reports. Both fail against the previous implementation.

The smaller items are done too: the misnamed test is now `TestReconcileMalformedDuplicatesDoNotTriggerDuplicateReport` and its body matches, and the `Reconcile` doc comment documents the `reportedDuplicates` lifecycle and marks the "new failure reason stays silent" behavior as deliberate. I checked all five doc bullets against the code — including the one that was inverted last round — and they're all true now. That was this PR's recurring weak spot, so it's good to see it closed.

I also checked the `GtfsRTFeed.AgencyIDs` removal, since it's outside the stated scope. It's safe: nothing reads the field, and there's no `DisallowUnknownFields` anywhere, so operator configs still carrying `agency_ids` keep decoding. I'd normally ask for that in its own PR, but it's tidying the same `compat.go` this change already rewrites, so I'll take it.

CI is green, the CLA is signed, and the full suite passes for me locally on a trial merge with #141.

Thanks for staying with this one through four rounds — the final shape is better than what any of the intermediate fixes would have given us. Merging.

Two follow-ups I'm filing rather than holding this on: `serverIdentityFromRaw`'s `duplicateEligible` return is now dead (the `counts` removal orphaned it), and the recovery report names the surviving valid entry rather than the one that was actually broken when a malformed and valid entry share an identity.

PR #131, opened 2026-08-13
Ship Watchdog’s New Architecture: Agency Scope and Multi-Agency Server Scope

0xaboomar:
## Summary by CodeRabbit

* **New Features**
  * Added agency-based active-vehicle monitoring with support for multiple GTFS and realtime feeds.
  * Added GTFS bundle last-fetched timestamps.
  * Added unresolved unmatched-stop and station-cluster tracking.
  * Automatically removes inactive unmatched-stop metrics after 24 hours.

* **Dashboard & Documentation**
  * Updated dashboards and metric documentation to use agency-based labels.
  * Added retention, historical-query, and alerting guidance.
  * Updated configuration examples for multiple GTFS and realtime feeds.

* **Bug Fixes**
  * Improved resilience when vehicle checks or stop lookups fail.
  * Sanitized server URLs in monitoring metrics to protect credentials and sensitive URL details.


aaronbrethorst:
### Code review

Found 2 issues:

1. The new/updated rows in `docs/METRICS.md` document labels that the code does not emit. This PR collapses every OBA metric to a single `agency_id` label, but the added rows say `gtfs_bundle_last_fetched_timestamp_seconds` is labeled `server` (L31), `oba_unmatched_stop_unresolved` is labeled `server`, `agency`, and `oba_unmatched_stop_cluster_count` is labeled `server`, `agency`, `station_id`, ... — while `internal/metrics/metrics.go` declares `[]string{"agency_id"}`, `[]string{"agency_id"}`, and `[]string{"agency_id", "station_id", "cluster_id", "cluster_lat", "cluster_lon"}` respectively. The copy-pasteable alert added at L119, `sum by (server) (oba_unmatched_stop_unresolved) > 0`, therefore has no `server` label to group on and collapses every agency into one nameless series. The rest of the section 5 table (`server`, `agency` on ten metrics) and the section 3 rows are also stale after this rename and are worth sweeping in the same pass, since CLAUDE.md points readers here as the full metric catalog.

https://github.com/OneBusAway/watchdog/blob/e048ff83402e152773eadb1cdeb77929901ca160/docs/METRICS.md#L96-L99

2. The Grafana dashboard still filters on the labels this PR deletes. The one query the PR rewrote keeps `server_id=~"$server_id"` on `oba_agency_active_vehicles_count`, which is now declared with only `agency_id`, so that selector can never match anything but the implicit "All". More broadly, the `server_id` template variable is defined as `label_values(server_id)` (L618) and no metric emits `server_id` any more, so the Server dropdown resolves to an empty option list; and panels 4.1/4.2/4.3 filter `agency=~"$agency_id"` on metrics that now carry `agency_id`, so picking a specific agency blanks them. `d95bf93 fix(grafana): correct server/agency label filtering on OBA metrics` fixed this same class of drift once already.

https://github.com/OneBusAway/watchdog/blob/e048ff83402e152773eadb1cdeb77929901ca160/grafana/dashboards/watchdog_metrics_dashboard.json#L272-L274

aaronbrethorst (changes requested):
This is a lot of careful work and most of it is good. The per-agency `RealtimeStore` is the standout — the single global slot on `main` genuinely mixes data across servers, and this fixes it. Locking on the shared stores is correct throughout, the label-set ordering in the new `UnmatchedStopTracker` lines up with `DeleteLabelValues` so series actually get pruned, and multi-bundle/multi-feed support is a real capability gain. CI is green.

The blockers are all in the observability surface rather than the Go code.

## 1. An API key is written to structured logs

`fetchObaAPIMetrics` turns the old `fmt.Printf` into:

```go
logger.Info("Fetching metrics from OBA server", "agency_id", agencyID, "url", url)
```

where `url` is `.../api/where/metrics.json?key=<apiKey>`. The `fmt.Printf` leaked it too, so this isn't new in substance — but this PR rewrote that exact line, and it adds `utils.SanitizeServerURL` for precisely this purpose. Please run the URL through the sanitizer here and in the `ExtraContext{"url": url}` in the same function. Now that it goes through `slog` rather than stdout, it lands wherever logs are aggregated.

## 2. `docs/METRICS.md` documents labels this PR doesn't emit

The rows added here list `server` / `server`, `agency`:

- `gtfs_bundle_last_fetched_timestamp_seconds` — documented `server`
- `oba_unmatched_stop_unresolved` — documented `server`, `agency`
- `oba_unmatched_stop_cluster_count` — documented `server`, `agency`, ...

but `metrics.go` in this same PR declares all three with `agency_id`. The example alert added alongside them, `sum by (server) (oba_unmatched_stop_unresolved) > 0`, groups on a label that doesn't exist, so it collapses to one aggregate instead of a series per server.

## 3. The dashboard filters on a deleted label

```
oba_agency_active_vehicles_count{server_id=~"$server_id", agency_id=~"$agency_id"}
```

`oba_agency_active_vehicles_count` is declared `[]string{"agency_id"}`. Nothing emits `server_id` any more, and the `server_id` template variable is `label_values(server_id)`, so it won't populate — that panel goes blank once a real value is selected. The panels filtering `agency=~"$agency_id"` have the same problem now that the label is `agency_id`.

## 4. Not a defect, but I want this to be a deliberate decision: it is a breaking config change

Dropping `id` and moving to `gtfs_urls[]` / `gtfs_rt_feeds[]` means every deployed `config.json` — and any payload served to `--config-url` — has to be migrated in the same window as the deploy. There is no compatibility shim, so an un-migrated config fails validation on every entry and monitoring goes dark. I'm fine with the schema; I want us to land it with a migration note in the README and a plan for the remote config, rather than discover it at deploy time.

## Smaller things, none blocking

- `CountActiveVehiclesForAgency` sits after the `FetchAndStoreGTFSRTFeed` gate but only hits the OBA API and never reads the realtime store, so `oba_agency_active_vehicles_count` stops updating exactly when the RT feed breaks. It could move above the gate — the `CLAUDE.md` line you updated describes the post-gate checks as the ones that read the realtime store, and this one doesn't.
- The OBA REST metrics are now labeled with the agency ID each server reports from its own `/metrics.json`. Those aren't globally unique across operators — `"1"` is common — so two monitored servers reporting the same agency ID overwrite each other every cycle. The new duplicate guard only covers the *configured* `agency_id`.
- `UnmatchedStopTracker.RecordLastSeen` freezes the first-seen `StopName`/`Lat`/`Lon`. If a stop is renamed or moves between bundle refreshes, the tracker prunes the old series and never tracks the new one.
- `GtfsRTFeed.AgencyIDs` isn't referenced anywhere in production code.
- Stale comments: `vehicle_metrics.go:107` still says `server_id`, and `CollectMetricsForServer`'s doc comment still lists the vehicle-count-match step this PR removes.

One sequencing note: this and #130 can't both merge as-is — #130's `DroppedServersStore` is keyed on the `server.ID` this PR deletes, and both rewrite `filterValidServers`. I'd like this one to land first since it's the larger structural change.

Happy to re-review. The code itself is in good shape; this is mostly getting the docs and dashboard caught up with the rename.

0xaboomar:
> 4. Not a defect, but I want this to be a deliberate decision: it is a breaking config change
> Dropping id and moving to gtfs_urls[] / gtfs_rt_feeds[] means every deployed config.json — and any payload served to --config-url — has to be migrated in the same window as the deploy. There is no compatibility shim, so an un-migrated config fails validation on every entry and monitoring goes dark. I'm fine with the schema; I want us to land it with a migration note in the README and a plan for the remote config, rather than discover it at deploy time.

Hey @aaronbrethorst, regarding this point I can add a fallback to ensure backward compatibility so we don’t break existing clients. Alternatively, we can explicitly document this in the README and the upcoming release notes. What do you prefer?

0xaboomar:
Hey @aaronbrethorst, I’ve added support for the legacy configuration as well to ensure backward compatibility. I decided to do this because the behavior wouldn’t change; we simply treat the legacy credentials as if the user configured one static feed and one realtime feed.

The changes to the metrics, which are now primarily filtered by `agencyId`, are still doable with the legacy configuration. We simply ignore the `id` and `serverId` fields.

So the behavior remains the same, and the end user gets the same experience. I’ll also mention in the next release notes that we’ve added support for multiple realtime and static feeds, while the legacy configuration is still supported. However, we now rely entirely on `agencyId` and no longer use `id` or `serverId`.

internal/gtfs/gtfs_bundles.go
```diff
@@ -39,43 +37,41 @@ import (
...
 	for _, server := range servers {
 		s := server
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-
-			staticBundle, err := downloadGTFSBundle(ctx, s.GtfsUrl, s.ID, maxRetries)
-			if err != nil {
+			bundles := make([]*remoteGtfs.Static, 0, len(s.GtfsStaticFeeds))
+			for _, gtfsURL := range s.GtfsStaticFeeds {
```
aaronbrethorst, line 66: One failing static feed discards every other feed for the agency.

Inside the goroutine, a download error `return`s outright, so `storeGTFSBundles` is never reached and the bundles already downloaded in this loop are thrown away. With a single feed this matches the old behavior, but with the new multi-feed support an agency configured with three `gtfs-static-feeds` where the third 500s ends up with **no** entry in `StaticStore` or `BoundingBoxStore` at all — so `checkBundleExpiration`, `getStopLocationsByIDs` (unmatched-stop resolution), and `trackInvalidVehiclesAndStoppedOutOfBounds` all fail for that agency until the next 24h refresh, and on a refresh it also leaves the previous good data in place only by accident (because `Set` is never called).

Consider logging/reporting the failed feed and `continue`ing, then storing whatever bundles succeeded (bailing only when `len(bundles) == 0`).

internal/gtfs/gtfs_bundles.go
```diff
@@ -271,55 +294,84 @@ func getStopLocationsByIDs(serverID int, stopIDs []string, staticStore *StaticSt
...
+			})
+			return err
+		}
+		for _, vehicle := range parsed.Vehicles {
+			id := ""
+			if vehicle.ID != nil {
+				id = vehicle.ID.ID
+			}
+			if id != "" {
+				if _, exists := vehicleIDs[id]; exists {
```
aaronbrethorst, line 638: Cross-feed vehicle-ID dedup silently drops distinct vehicles, and `GtfsRTFeed.AgencyIDs` is ignored.

`vehicleIDs` is shared across every feed in `server.GtfsRTFeeds`, so the first feed wins for any repeated ID. GTFS-RT vehicle IDs are only unique within a feed/agency — two feeds that both contain vehicle `"101"` (very common with small integer vehicle IDs) will have the second one dropped, undercounting `realtime_vehicle_positions_count_gtfs_rt` and `gtfs_rt_tracked_vehicles_count`.

The flip side is also unhandled: every vehicle from every configured feed is merged into the single `server.AgencyID` bucket, and `feed.AgencyIDs` — which the config template tells users to fill in — is never consulted to filter. A feed covering several agencies inflates the configured agencys vehicle count, out-of-bounds count, and per-vehicle telemetry series with vehicles belonging to other agencies.

internal/config/compat.go
```diff
@@ -0,0 +1,164 @@
...
+	for _, raw := range rawEntries {
+		server, err := decodeServerEntry(raw)
+		if err != nil {
+			report.ReportErrorWithSentryOptions(err, report.SentryReportOptions{
+				Tags:  serverTagsFromRaw(raw),
+				Level: sentry.LevelError,
+			})
+			continue
+		}
+		if _, exists := seenAgencies[server.AgencyID]; exists {
```
aaronbrethorst: Duplicate `agency_id` silently drops a whole server, and agency IDs are not globally unique.

This PR removes the integer `id` and keys everything (`StaticStore`, `BoundingBoxStore`, `BackoffStore`, and every metric label) on `agency_id` alone. GTFS `agency_id` is only unique *within* one OBA server — values like `"1"` or `"MTA"` are reused across regions. A fleet config that monitors two OBA deployments which both expose `agency_id: "1"` will now have the second entry dropped here entirely; it also silently unifies their stores if the dedup were removed.

Two concerns:
1. The drop is reported only to Sentry via `report.ReportError` — there is no `logger.Error`, unlike every other config failure path, so an operator running locally sees a server just vanish with no output.
2. Consider keying the stores on something guaranteed-unique (e.g. `oba_base_url` + `agency_id`) so distinct servers sharing an agency ID can both be monitored.

internal/models/oba_server.go
```diff
@@ -1,32 +1,26 @@
...
-	TripUpdateUrl      string `json:"trip_update_url"`
-	VehiclePositionUrl string `json:"vehicle_position_url"`
-	GtfsRtApiKey       string `json:"gtfs_rt_api_key"`
-	GtfsRtApiValue     string `json:"gtfs_rt_api_value"`
-	AgencyID           string `json:"agency_id"`
+	AgencyName      string       `json:"agency_name"`
+	AgencyID        string       `json:"agency_id"`
+	ObaBaseURL      string       `json:"oba_base_url"`
+	ObaApiKey       string       `json:"oba_api_key"`
+	GtfsStaticFeeds []string     `json:"gtfs-static-feeds"`
```
aaronbrethorst: The JSON key `gtfs-static-feeds` is hyphenated while every other key in this struct (and in `GtfsRTFeed`) is snake_case.

A config author who writes the consistent-looking `gtfs_static_feeds` gets no unmarshal error — the field stays nil, `ValidateServer` reports `missing required fields: gtfs-static-feeds`, and `decodeServers` drops the entry to Sentry. `compat.go` also keys its v1/v2 discrimination off `fields["gtfs-static-feeds"]`, so the typo silently falls through the v2 branch. Suggest renaming to `gtfs_static_feeds` (or accepting both) while the schema is still new.

internal/models/oba_server.go
```diff
@@ -1,32 +1,26 @@
...
-		AgencyID:           agencyID,
-	}
+// GtfsRTFeed is one GTFS-Realtime source. Trip updates are retained for future
+// consumers; current monitoring uses vehicle positions.
+type GtfsRTFeed struct {
+	TripUpdateURL      string   `json:"trip_update_url"`
+	VehiclePositionURL string   `json:"vehicle_position_url"`
+	GtfsRTAPIKey       string   `json:"gtfs_rt_api_key"`
+	GtfsRTAPIValue     string   `json:"gtfs_rt_api_value"`
+	AgencyIDs          []string `json:"agency_ids"`
```
aaronbrethorst, line 46: `AgencyIDs` is never read anywhere in the codebase.

`config.json.template`, `integration_config.json.template`, and the README all instruct users to populate `agency_ids` on each RT feed, but nothing consumes it — `fetchAndStoreGTFSRTFeed` merges all feeds into the single `server.AgencyID` bucket regardless. Either wire it up (filter merged vehicles by the feeds declared agencies) or drop it from the model and the templates, otherwise operators will configure it and reasonably expect per-agency attribution they are not getting.

internal/metrics/oba_rest_api_metrics.go
```diff
@@ -87,108 +99,164 @@ func fetchObaAPIMetrics(slugID string, serverID int, serverBaseUrl string, apiKe
...
-			Tags: utils.MakeMap("slug_id", slugID),
+			Tags: utils.MakeMap("agency_id", agencyID),
 			ExtraContext: map[string]interface{}{
-				"url":         url,
+				"url":         sanitizedURL,
 				"status_code": resp.StatusCode,
 			},
 		})
 
+		setObaApiMetricsStatus(agencyID, agencyName, serverBaseUrl, false)
```
aaronbrethorst: 404 from `/metrics.json` now permanently pins `oba_api_status` to 0 for servers that simply do not have the endpoint.

The 404 branch explicitly recognizes "server does not support metrics API" (OBA < 2.6.0), yet still calls `setObaApiMetricsStatus(..., false)`. On `main`, `fetchObaAPIMetrics` only ever set the gauge to 1; it never set 0. So this is a new behavior: for a healthy pre-2.6.0 server, `serverPing` writes `oba_api_status{server_url=".../current-time.json"} 1` and this writes `oba_api_status{server_url=".../metrics.json"} 0` every tick. The dashboard panel queries `oba_api_status{agency_id=~...}` without filtering on `server_url`, so the agency renders as down forever.

Suggest leaving the gauge untouched (or using a distinct "metrics API supported" series) on 404, and reserving 0 for genuine failures.

internal/metrics/metrics.go
```diff
@@ -12,195 +12,193 @@ var (
...
-	}, []string{"server", "agency"})
+	}, []string{"agency_id", "agency_name"})
 
 	ObaTimeSinceUpdate = promauto.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "oba_time_since_last_update_seconds",
 			Help: "Time since last realtime update in seconds",
 		},
-		[]string{"server", "agency"},
+		[]string{"agency_id", "agency_name"},
```
aaronbrethorst: The `server` label is gone, but the alert rules on `main` still interpolate it.

`prometheus/rules/feed_staleness.yml` and `prometheus/rules/prediction_quality.yml` (added in #129, merged after this branch was cut) alert on `oba_time_since_last_update_seconds`, `oba_realtime_trip_match_ratio`, and `oba_stop_match_ratio`, and their annotations read `{{ $labels.server }}`. After this rename every one of those alerts will render "GTFS-RT feed stale for " / "for server  is ..." with an empty value, since the label is now `agency_name`. Please update the three rule files as part of this PR.

internal/metrics/tracked_agencies.go
```diff
@@ -0,0 +1,44 @@
...
+// It reports the total count to the AgenciesTrackedCount Prometheus metric,
+// and emits one AgenciesTrackedInfo series per agency labeled with its ID,
+// name, and base URL.
+//
+// This is called once at startup and again only when the configured server set
+// changes (e.g., a remote config refresh adds or removes an agency), never on
+// the periodic metric collection tick. Re-reporting an unchanged set is a no-op.
+func reportTrackedAgencies(servers []models.ObaServer) {
+	ids := trackedAgencyIDs(servers)
+	if trackedAgenciesReported && sameAgencySet(ids, lastTrackedAgencies) {
```
aaronbrethorst: The short-circuit compares only the agency ID *set*, so label changes never propagate.

If a remote config refresh renames an agency or moves it to a new `oba_base_url` without adding or removing any agency, `sameAgencySet` returns true and the function returns early. `oba_tracked_agencies_info` then keeps serving the stale `agency_name` / `server_url` labels for the lifetime of the process. Comparing the full (id, name, url) tuples instead of just the IDs would fix it and costs nothing at this call frequency.

internal/metrics/unmatched_stop_tracker.go
```diff
@@ -0,0 +1,195 @@
...
+	defer t.Mu.Unlock()
+
+	stops, ok := t.Entries[agencyID]
+	if !ok {
+		stops = make(map[string]trackedStop)
+		t.Entries[agencyID] = stops
+	}
+
+	entry, exists := stops[stopID]
+	if exists && (entry.StopName != stopName || entry.Lat != lat || entry.Lon != lon) {
```
aaronbrethorst: The staleness check omits `agencyName`, so a renamed agency leaks its old series forever.

The guard deletes the previous series when the stop name or coordinates change, but not when `agencyName` differs from `entry.AgencyName`. The entry is then overwritten with the new name, so both this delete path and `clearStops` will only ever delete the series labeled with the *new* name — the `oba_unmatched_stop_info{agency_name="<old>"}` series is never removed, even after the 24h TTL. Add `entry.AgencyName != agencyName` to the condition.

aaronbrethorst (changes requested):
Thanks for taking this on — the `server_id` → `agency_id`/`agency_name` rename is the right direction, and the multi-feed config, SDK client cache, and URL sanitization are all welcome. The branch builds clean, `go vet` is quiet, and the tests pass. I've left nine inline comments; a handful need to be resolved before this can land.

**Blocking**

1. **Alerting rules break on merge.** This branch predates #129, which added `prometheus/rules/feed_staleness.yml` and `prediction_quality.yml`. Those rules interpolate `{{ $labels.server }}`, and this PR removes the `server` label — so after merge, three alerts render with empty agency names. Please rebase on `main` and update the rule files here.

2. **`downloadGTFSBundles` aborts the whole agency on one bad feed** (`internal/gtfs/gtfs_bundles.go:53`). It `return`s on the first failing static feed, throwing away bundles already downloaded and skipping `storeGTFSBundles` entirely. A three-feed agency with one flaky URL ends up with no `StaticStore`/`BoundingBoxStore` entry at all — for 24h, until the next refresh. Log-and-continue per feed, and store whatever succeeded.

3. **RT vehicle dedup is cross-feed** (`internal/gtfs/gtfs_bundles.go:366`). Vehicle IDs are only unique within a feed, so two feeds that both report vehicle `"101"` silently lose one. The merge also dumps every feed into a single agency bucket and ignores `feed.AgencyIDs`.

4. **Duplicate `agency_id` silently drops a server** (`internal/config/compat.go:156`). It goes to Sentry with no log line. Agency IDs are only unique per OBA server, so two regions that both use `"1"` will collide — and now that every store and the backoff state key off `agency_id` alone, that collision is much easier to hit than it looks. At minimum this needs a loud log; ideally the key stays scoped to the server.

5. **404 from `/metrics.json` now reports the server as down** (`internal/metrics/oba_rest_api_metrics.go:112`). Pre-2.6.0 servers 404 here — the branch recognizes that case a few lines up — but the new code sets `oba_api_status` to 0, which `main` never did at this call site. Healthy older servers will show as permanently down, and the dashboard doesn't filter on `server_url`.

**Non-blocking, but worth fixing while you're in here**

6. `gtfs-static-feeds` (`internal/models/oba_server.go:9`) is hyphenated while every other key is snake_case. The natural `gtfs_static_feeds` typo unmarshals to nil and the entry disappears without complaint — and `compat.go` discriminates v1 vs v2 configs off that exact name.

7. `GtfsRTFeed.AgencyIDs` (`internal/models/oba_server.go:20`) is never read anywhere, yet the templates and README instruct users to populate it. Either wire it up (see #3) or drop it from the docs.

8. `sameAgencySet` (`internal/metrics/tracked_agencies.go:28`) compares only IDs, so a name or base-URL change in a remote config never refreshes `oba_tracked_agencies_info` labels for the life of the process.

9. The stale-series check in `internal/metrics/unmatched_stop_tracker.go:76` omits `agencyName`, so after a rename the old `oba_unmatched_stop_info{agency_name="<old>"}` series is never cleaned up, even past the 24h TTL.

Happy to look again once the rebase and the first five are addressed.

internal/config/compat.go
0xaboomar: Hey @aaronbrethorst , Should I bring back the serverId and use {serverId} - {agencyId} as the key in stores ?

internal/models/oba_server.go
0xaboomar: Okay, I’ll drop it for now and we can consider adding it in the future if we need to filter the data based on `agency_ids`.

internal/metrics/oba_rest_api_metrics.go
0xaboomar: No worries! This case was already addressed in a commit made a few minutes before this review.

internal/metrics/metrics.go
0xaboomar: I will rebase the branch with main and resolve this issue.

internal/config/compat.go
0xaboomar: I will use `oba_base_url` + `agency_id` as you suggested.

0xaboomar:
Hey @aaronbrethorst, the point you raised here is really important.

> The flip side is also unhandled: every vehicle from every configured feed is merged into the single server.AgencyID bucket, and feed.AgencyIDs — which the config template tells users to fill in — is never consulted to filter. A feed covering several agencies inflates the configured agency's vehicle count, out-of-bounds count, and per-vehicle telemetry series with vehicles belonging to other agencies.

I've deliberately designed it this way because an agency can provide multiple static and realtime feeds, and I treat all of those feeds as a single combined feed belonging to that configured umbrella agency.

Because of that, I don't use the `AgencyIDs` list from the config to filter vehicles within each feed. Regardless of which agency in that list owns or contributes data to a particular feed, I treat the feed as part of the configured umbrella agency's combined dataset.

Similarly, I only generate metrics for the umbrella agency defined in the config. There is no additional filtering or separate metric generation for the individual agencies in `AgencyIDs`, since the multiple feeds are intentionally treated as one combined feed owned by the umbrella agency.

Is there anything I should revise or change in this approach?

internal/models/oba_server.go
0xaboomar: Hey @aaronbrethorst, I changed my mind on this. I think we should keep it even though we don't use it currently. We may want to add filtering logic based on `agency_ids` or use it for another use case in the future. Keeping it now avoids potentially breaking clients later if we decide to use it.

aaronbrethorst:
## Review + fixes applied

Mohamed — this is a genuinely impressive piece of work. Reshaping the whole system from `server_id`-keyed state to composite `(oba_base_url, agency_id)` keys, introducing the scope abstraction, and keeping the v1 config schema working through `compat.go` is a lot of surface area to move at once, and the care shows: the pointer-shared bundle storage, the `O(1)`-RT-fetch design in server-mode, the collision detection in `mergeStaticAndDiscoverAgencies`, and the honest `TODO(scoped-store)` notes that say what's deliberately deferred and why. The comments explaining *why* rather than *what* made this a pleasure to review. Thank you for the diligence.

I found a handful of real bugs while reading it and pushed fixes. Full build/vet/test pass after the changes. Details below so you can push back on anything you disagree with.

### Fixed

**1. Agency-mode never fetched the GTFS-RT feed (critical).** `CollectMetricsForServer` no longer calls `FetchAndStoreGTFSRTFeed`, and its doc says "agency-mode callers invoke it before calling this method" — but `collectForScope`'s `AgencyScope` branch calls `CollectMetricsForServer` directly. `FetchAndStoreGTFSRTFeed` had zero production callers on the branch. Every agency-scoped entry would have logged `no GTFS-RT data available for agency X` on every tick forever, and `realtime_vehicle_positions_count_gtfs_rt`, `gtfs_rt_tracked_vehicles_count`, `vehicle_report_total`, `gtfs_rt_invalid_vehicle_coordinates` and `gtfs_rt_stopped_out_of_bounds_vehicles` would never have been populated.

Fix: split into an exported `CollectMetricsForServer` (fetches RT) and an unexported `collectMetricsForServer(server, fetchRealtime bool)`. Server-mode passes `false` so it keeps its one-fetch-per-tick optimisation. The RT fetch stays a hard gate — a failure returns before the vehicle metrics, matching the pre-PR ordering. Added `TestAgencyScopeFetchesRealtimeFeed`, which I verified fails without the fix.

**2. Agency-mode bundles were stored under the wrong key.** `storeStaticForServer` keys the bundle purely off `agency.txt`, ignoring the configured `server.AgencyID` — even though its own doc says "Agency-mode: the bundle is stored under `server.ServerKey()` exactly once." Every agency-mode reader (`checkBundleExpiration`, `getStopLocationsByIDs`, the bbox lookup) derives its key from `server.ServerKey()`. So whenever the feed's `agency_id` differs from the configured one, nothing is ever found. And a single-agency feed may legally omit `agency_id` entirely, in which case `declaredAgencies` is empty and the bundle is dropped altogether. `TestStoreGTFSBundleRecordsFetchTime` had actually been written *around* this — it looked the fixture's real agency id ("40") up rather than using the configured `agency-1`, with a comment describing the behaviour as intended.

Fix: in agency-mode, store once under `server.ServerKey()`, borrowing `agency_name` from `agency.txt` when it's available there. Server-mode is unchanged. Updated that test to assert the documented behaviour and added a server-mode counterpart.

**3. Server-mode backoff was write-only.** `collectForServerScope` calls `UpdateBackoff`/`ResetBackoff` on `server.ServerKey()`, but nothing ever calls `NextRetryAt` for that key — the per-agency check inside `CollectMetricsForServer` uses per-agency keys. A dead server got re-pinged every tick with the backoff state accumulating unread. Added the `NextRetryAt` check at the top of `collectForServerScope`.

**4. `oba_api_status`'s `server_url` couldn't join with anything.** `serverPing` labelled it `SanitizeServerURL(ObaBaseURL + "/api/where/current-time.json")`, while every other metric — and the dashboard's `$server_url` variable, sourced from `oba_tracked_agencies_info` — uses the bare sanitized base URL. Changed to the bare base URL and updated `server_ping_test.go`.

**5. Dashboard panel filtered on a label that no longer exists.** `oba_api_status{agency_id=~"$agency_id"}` — the metric's labels are now `server_name`/`server_url` only, so the "Is the API up?" panel would always render empty. Switched to `server_url=~"$server_url"` (which works now that #4 is fixed).

**6. Two server-scope gauges leaked the raw base URL.** `GtfsStaticAgencyCurrentlyLive` and `GtfsStaticFeedAttributionStatus` were labelled with `server.ObaBaseURL` and `feedURL` verbatim. Everything else runs through `SanitizeServerURL` — which exists precisely so credentials embedded in a URL never reach a label — and the raw form also wouldn't match the dashboard's `$server_url`. Both now sanitized.

**7. A test assertion that could never fail.** `TestFetchObaAPIMetrics_SetsStatusZeroOnFailure` asserted `oba_api_status == 0`, but `fetchObaAPIMetrics` no longer writes that gauge, and `getMetricValue` uses `metric.With(labels)`, which *creates* the series on read. So it read back a freshly-minted zero and passed unconditionally. Rewrote it to assert that a failed call emits no per-agency series, via a new non-mutating `countSeries` helper. Worth keeping in mind for other tests using `getMetricValue`.

**8. `trackInvalidVehiclesAndStoppedOutOfBounds` doc didn't match the code.** It took a `routeAgencyIndex` parameter, never used it, and its doc claimed "it attributes vehicles to agencies via the RouteAgencyIndex in server-mode." Dropped the unused parameter and replaced the doc with what actually happens, including the server-mode caveat (see #9).

**9. Smaller cleanups.** Removed the dead unexported `calculateNextRetryAt`/`calculateNewBackoffDelay` duplicates in `utils/backoff.go`; replaced the `metricsResponse` struct in `metrics_collector.go` with the existing `metrics.OBAMetrics` so the two decoders of `/metrics.json` can't drift; stopped `legacyToCurrent` emitting `AgencyIDs: []string{""}` for legacy entries with no `agency_id`; corrected the `RouteAgencyIndex.Set` doc, which claims the index is keyed by the sanitized base URL when it's actually stored verbatim (behaviour is self-consistent today, but the comment invites a future mismatch).

### Not fixed — needs your call

**Server-mode double-counting in `trackVehicleTelemetry`.** In server-mode the per-agency loop calls this once per live agency over the *same* merged RT feed. `VehicleReportCount` is a Counter, so every vehicle gets `Inc()` N times per tick on an N-agency server — `rate()` comes out N× too high. Relatedly, `vehicleLastSeen.Set(server.ServerKey(), ...)` records every vehicle under *every* agency's key, so `gtfs_rt_tracked_vehicles_count` reports the whole server's fleet for each agency. `countVehiclePositions` and `trackInvalidVehiclesAndStoppedOutOfBounds` have the same shape (each agency reports server-wide totals).

This is the same underlying issue as your `TODO(scoped-store)` in `config/scoping.go` and the bbox `NOTE` in `storeStaticForServer` — the merged-bundle storage shape loses the agency↔feed relationship. It needs a design decision rather than a review-time patch, so I left it alone. The counter inflation is the part I'd prioritise, since gauges self-correct on the next scrape but a counter never does.

**Stale keys are never pruned.** `staticStore`, `boundingBoxStore` and `RealtimeStore` accumulate entries per serverKey and nothing removes them when an agency disappears from `agency.txt` or from the remote config. `discoverAgenciesForServer` will keep returning the departed agency indefinitely. Slow leak, not urgent, but worth a follow-up issue.

PR #136, opened 2026-08-26
docs: watchdog server scoped config

0xaboomar:
## Summary by CodeRabbit

- **Documentation**
  - Added guidance for agency-level and server-wide monitoring modes.
  - Documented configuration scope, agency discovery, liveness monitoring, and vehicle attribution.
  - Updated examples to demonstrate server-wide monitoring with feed-level agency settings.
  - Retained v1-to-v2 backward-compatibility guidance and removed duplicate documentation.

- **Configuration**
  - Clarified example names for agency and server configurations.
  - Updated the server-wide example to use an empty agency list.


aaronbrethorst:
### Code review

Found 2 issues:

1. "data for any other agency is ignored" (and, at line 87, "any data not related to it is ignored") is not what agency mode does. In agency mode `collectVehicleMetrics` is called with `nil` agencies, and `attributeVehicle` then returns the configured entry for *every* vehicle in the feed (`internal/metrics/vehicle_metrics.go`: `if agencyByID == nil { return server, true }`) — nothing is filtered. Likewise `storeStaticForServer` stores the *merged* bundle under the single configured agency key in agency mode (`internal/gtfs/gtfs_bundles.go:127-136`), so a multi-agency static feed's stops/routes are all counted under that one agency. This README says so itself two sections down: "In agency-mode every RT vehicle is labeled with the configured `agency_id`." Only the OBA REST probes (`CountActiveVehiclesForAgency`, `FetchObaAPIMetrics`) are genuinely agency-filtered.

https://github.com/OneBusAway/watchdog/blob/52faafa2d6e3ef372d5ed8e2757db568f745b462/README.md#L43-L45

2. Renaming the section from "Server vs. agency scoping" to "Two observation modes: agency vs. server" leaves a dangling cross-reference in the code: `cmd/watchdog/main.go:30-31` reads `// Server-scope design (deliberate decision, see README "Server vs. agency scoping" and config.json.template)`, and that heading no longer exists anywhere in the repo after this change.

https://github.com/OneBusAway/watchdog/blob/52faafa2d6e3ef372d5ed8e2757db568f745b462/README.md#L79-L81

aaronbrethorst (changes requested):
The restructuring here is a real improvement. The two-mode framing with a worked example for each plus the comparison table is much easier to follow than what it replaces, and moving backward compatibility below it is the right order. I checked the substantive claims against the code and most hold up: the `/api/where/metrics.json` cross-referenced with `agency.txt` description is accurate, the table's split is right, and your note that feed-level `agency_ids` is "optional and unused by Watchdog today" is correct — `models.GtfsRtFeed.AgencyIDs` is only ever written by the v1-to-v2 shim in `internal/config/compat.go` and never read in production. Good catch documenting that.

Two things to fix.

**"data for any other agency is ignored" isn't true.** It appears twice, in the new paragraph near the top and again in the agency-mode section. In agency mode `attributeVehicle` returns the configured entry for every vehicle it sees — `agencyByID == nil` short-circuits before any route lookup — so every vehicle in a merged RT feed gets labeled with the configured `agency_id` regardless of which agency it belongs to. `storeStaticForServer` likewise stores the *merged* bundle under the single configured agency key, so static stop and route counts include every agency in the feed. Only the OBA REST probes are genuinely agency-filtered. As written this tells an operator that cross-agency contamination is impossible in agency mode, which is the exact problem server mode was built to solve, and the README's own vehicle-attribution paragraph says the opposite a few sections down. Something like "the OBA API probes query only that agency, though static and realtime feed data is not filtered" would be accurate.

**The heading rename orphans a pointer in the code.** `cmd/watchdog/main.go:30` says `see README "Server vs. agency scoping"`, and after this PR no such heading exists. Please update that comment to the new heading in the same PR.

Two small ones: `config.json.template` loses its trailing newline, and the v1 mapping bullet you relocated says `name` maps to `agency_name`, which doesn't match `legacyToCurrent` — `name` always becomes `server_name`, and `agency_name` is only set when `agency_id` is present. That bullet was already wrong on main so it isn't something you introduced, but it'd be good to correct since you're touching the block.

Fix the two claims and I'll merge this.

0xaboomar:
> "data for any other agency is ignored" isn't true. It appears twice, in the new paragraph near the top and again in the agency-mode section. In agency mode attributeVehicle returns the configured entry for every vehicle it sees — agencyByID == nil short-circuits before any route lookup — so every vehicle in a merged RT feed gets labeled with the configured agency_id regardless of which agency it belongs to. storeStaticForServer likewise stores the merged bundle under the single configured agency key, so static stop and route counts include every agency in the feed. Only the OBA REST probes are genuinely agency-filtered. As written this tells an operator that cross-agency contamination is impossible in agency mode, which is the exact problem server mode was built to solve, and the README's own vehicle-attribution paragraph says the opposite a few sections down. Something like "the OBA API probes query only that agency, though static and realtime feed data is not filtered" would be accurate.


This is not a documentation-level problem. It identifies a behavior issue in agency mode.
Currently, when an entry has an agency_id, Watchdog assumes that every vehicle in the merged GTFS-RT feed belongs to that configured agency and labels all vehicles accordingly. The merged static bundle is also stored under that agency’s key, so static stop and route counts can include data from other agencies. Only the OBA REST API metrics and the VehiclesForAgency request are actually filtered by agency.
The intended solution is to scope both GTFS-realtime and GTFS static processing to the configured agency in agency mode, while keeping server-mode route-based attribution across agencies.
I have opened a separate issue #146  for this. Part of the implementation would require merging https://github.com/OneBusAway/watchdog/pull/141, which resolves the loose bounding-box calculation and would help with the agency-specific bounding-box handling here.

aaronbrethorst (changes requested):
Four of the five items from last time are cleanly resolved, and I checked each against the code:

- The `main.go:30` pointer now reads `see README "Two observation modes: agency vs. server"`, and that heading exists. I grepped the tree for the old heading and there are no other danglers.
- The v1 mapping bullet is correct now: `name` → `server_name`, with `agency_name` only when `agency_id` is present. That matches `legacyToCurrent` in `internal/config/compat.go`.
- `config.json.template` has its trailing newline back.
- The `agency_ids` "optional and unused by Watchdog today" note holds up — `AgencyIDs` is written only by the v1 shim and never read in production.

The restructure itself is good. The two-mode framing with a worked example each, the comparison table, and moving backward compatibility below it all make this much easier to follow than what it replaces.

But the one item I said I'd merge on is untouched, so I can't merge it yet.

Both sentences are still there at `f044d152`:

- README line 44: "set it to track only that agency (data for any other agency is ignored)"
- README line 87: "Watchdog tracks only that agency; any data not related to it is ignored"

I re-checked the code before writing this rather than relying on my earlier read. `attributeVehicle` still opens with:

```go
if agencyByID == nil {
    return server, true
}
```

and agency-mode calls it with `nil`, so every vehicle in the merged feed is attributed to the configured agency with no route lookup at all. `storeStaticForServer` still stores the merged bundle under the single configured agency key, so the static stop and route counts include every agency in the feed. Only the OBA REST probes actually filter by agency.

I understand the reasoning in your reply, and filing #146 is the right long-term move — the underlying behavior probably should change. But that's a separate argument from what this PR does, which is describe the system as it exists today. As written, this tells an operator that agency mode gives them cross-agency isolation, and it doesn't. That's the exact problem server mode was built to solve, and README line 147 in this same file says the opposite a few sections down.

This is a wording change, not a code change, and it doesn't depend on #146 or #141:

> the OBA API probes query only that agency, though static and realtime feed data is not filtered

Two small ones while you're in there, neither blocking:

- Moving the h3 down put `#### Backward compatibility` and `### Backward Compatibility (v1 → v2)` adjacent, and they now state the same `name` → `server_name` rule back to back. Worth merging them.
- Line 82 says `agency_id` is "the only switch", but `config_validation.go` rejects an entry with `agency_id` set and `agency_name` blank. The table two lines down gets it right, so it corrects itself quickly.

Separately, the CLA is blocking this and your four other open PRs. One of two committers has signed; the unsigned commits are authored as `Mohamed Ahmed Aboomar <aboomar@Mohameds-MacBook-Air.local>`, a local machine hostname rather than a real address, so the bot's suggestion to add that email to your GitHub account won't work. Rewrite the authorship and force-push:

```
git push --force-with-lease
```

Reword those two sentences and this is ready.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the one I said I'd merge on, and it's fixed. Both sentences now read the way the code actually behaves:

- "OBA API probes query only that agency, though static and realtime feed data is not filtered by agency"
- "Watchdog uses that agency for agency-specific OBA API probes, but static and realtime feed data is not filtered by agency"

I re-checked the underlying claim rather than assuming: `attributeVehicle` still short-circuits on `agencyByID == nil` in agency mode, and `storeStaticForServer` still stores the merged bundle under the single configured agency key. So the corrected wording is accurate and the README no longer promises isolation that agency mode doesn't provide. Thanks for making the change even though you disagreed with it — filing #146 for the underlying behavior is the right split, and this PR describing today's system is the part I needed.

The other four items from the earlier rounds all hold up: the `main.go:30` pointer resolves to a heading that exists, the v1 mapping bullet matches `legacyToCurrent`, the template's trailing newline is back, and the `agency_ids` note is correct.

The restructure is a genuine improvement — the two-mode framing with a worked example each, the comparison table, and pushing backward compatibility below it are all much easier to follow than what they replace.

Merging.

Issue #140, opened 2026-08-29
stop_id collisions silently discard a physical stop location


0xaboomar:
### Problem

When two configured static feeds contain the same `stop_id` but assign it different latitude/longitude coordinates, `mergeStaticAndDiscoverAgencies` (`internal/gtfs/gtfs_bundles.go:264-293`) keeps the first occurrence and silently discards the second location.

Currently, only a Sentry warning is emitted.

This means the second physical stop becomes completely invisible to the system. In particular, its location cannot be used by:

- Bounding-box validation. 
- Stop resolution.
- Other logic that relies on the merged stop lookup.
    

### Current constraint

Until this is fixed, **multi-feed configurations must not contain `stop_id` collisions**.

In practice, configurations should use a static feed bundle that has already been merged outside Watchdog rather than relying on Watchdog to combine multiple per-agency feeds that may contain overlapping `stop_id`s.

### Proposed fix

Instead of discarding the duplicate location, preserve the conflicting locations and make them available to the relevant processing paths.

The bbox-check and unmatched-stop resolution paths should then operate with the appropriate location and expose conflicts through per-`(agency, stop_id)` metrics where necessary.

This issue is also closely related to #137 : once bounding boxes become agency-specific, preserving the correct physical location for a colliding `stop_id` becomes even more important. A union-based stop lookup can only return one location for a duplicated ID, which can lead to incorrect agency-level bounding-box and stop-resolution behavior.

yanaix10:
Hey @0xaboomar  can you please assign this to me so that I can work on it.

0xaboomar:
Hey @yanaix10, thanks for your interest! This issue requires quite a bit of domain knowledge, as we need to handle collisions the same way our backend system does. I’m currently working on it, and I’ll open a good introductory issue for you to work on.

yanaix10:
okay thanks @0xaboomar  for the time being if you have any beginner friendly issue anywhere please ping me there I would be happy to work on that. [webdev , microservices ] any.

PR #141, opened 2026-08-29
fix(gtfs): handle multi-feed stop collisions in metrics and bounding boxes

### LINKED ISSUE #137: Per-agency bounding box is too broad in server-mode
### Problem

In server-mode, one OBA server can serve multiple agencies. Currently, `storeStaticForServer` (`internal/gtfs/gtfs_bundles.go`) merges all static feeds into one bundle and computes **one bounding box from all stops across all agencies**.

That same bounding box is then used when validating vehicles.

This means that a vehicle belonging to Agency A can be outside Agency A's service area but still be inside Agency B's service area, and therefore **will not be reported as out-of-bounds**.

Each agency should instead be validated against the area covered by its own feeds.

In other words, Agency A's vehicles are currently validated against **A ∪ B ∪ C**, rather than against the area covered by Agency A's feeds.

### Possible solutions

There are two ways to achieve this:

**1. Keep one shared bundle and compute a bounding box per agency**

All agencies continue pointing to the **same merged ****`*StaticData`**** bundle**, so the static data is loaded only once.

During the merge/storage step, we compute a separate bounding box for each agency using only the feeds associated with that agency. The per-agency bounding boxes are then stored alongside the shared bundle.

When validating vehicles, we may still scan the larger shared bundle for each agency, but this is acceptable because it avoids duplicating the static data in memory.

**2. Split the bundle per agency**

Instead of keeping one merged bundle, create a separate bundle for each agency containing only the feeds that belong to that agency. Each agency would then point to its own bundle, and its bounding box would be computed directly from that bundle.

This would make both the data and bounding box fully scoped to the agency, but it would also duplicate static data between agencies.

### Decision

At first glance, the second solution may seem better because it only loads the data needed by each agency, which sounds like it would save memory. In practice, the opposite is likely to happen over time because the data is loaded repeatedly every 30 seconds, resulting in duplicated data remaining in memory.

So **We should go with solution 1: keep one shared bundle and compute a bounding box per agency. Do not split the bundle.**

The main reason is memory usage over time.

Metrics are collected every 30 seconds. With solution 2, each agency would have its own copy of the relevant static data. Even though each individual validation only needs one agency's bundle, **over a 30-second collection window all of those bundles are loaded into memory**.

If the garbage collector does not reclaim the redundant data aggressively enough, those duplicated bundles can remain in memory for the entire 30-second window. Since the same process repeats every 30 seconds throughout the day, this can result in the redundant data remaining resident and consuming significantly more memory.

With solution 1, there is only **one copy of the merged static data**. We accept the additional RT scanning cost because the validation is not heavy and this solution saves are more important.

The implementation should therefore:

* Keep a single shared `*StaticData` bundle for all agencies.
* Compute a separate bounding box for each agency from the feeds associated with that agency.
* Store the per-agency bounding boxes alongside the shared bundle.
* Use the appropriate agency's bounding box when validating its vehicles.

The main complication is that the current storage does not preserve which feeds are associated with which agencies. The implementation will need to retain that association during the merge/storage process so that the correct bounding box can be computed for each agency.

0xaboomar:
closes #137 and #140  

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vehicle location validation using attributed agency boundaries, with server-wide fallback when needed.
  * Stop lookups now consistently use the first occurrence from the merged GTFS data.
  * Unmatched stop tracking distinguishes physical locations and handles renamed or relocated stops.
  * Updated unmatched-stop metrics and clustering to reflect resolved physical stop locations.

* **Documentation**
  * Clarified merged-stop resolution and physical-location metric behavior.

* **Tests**
  * Added coverage for agency boundaries, fallback behavior, duplicate stop IDs, and stop relocations.


aaronbrethorst:
### Code review

Found 3 issues:

1. The only production caller now uses `RecordLocationLastSeen`, which never retires a stale label set, so a renamed or relocated unmatched stop leaves its old `oba_unmatched_stop_info` series on `/metrics` for up to 24h alongside the new one — `sum by (agency_id, stop_id) (max_over_time(oba_unmatched_stop_info[1d]))` double-counts it for that window. That is exactly what `438dab7 fix(metrics): track renamed unmatched stops by latest labels` and `30fca50 fix(metrics): delete unmatched-stop series when agency name changes` ("so the old series is pruned on the first observation under the new label") were added to prevent. After this change `RecordLastSeen` has no production caller at all — the rename path is dead code, and `TestRecordLastSeenUpdatesLabelsOnRename` still passes because it calls the dead method directly, so CI can't see the regression. Consider recording the whole resolved location set for a stop ID in one call so locations absent this tick are retired while colliding ones coexist.

https://github.com/OneBusAway/watchdog/blob/bd30b8d3341211f3c44b7f13a8bfcaf578661cf9/internal/metrics/oba_rest_api_metrics.go#L253-L257

2. `allStops` builds the server-wide union box only from `StopsByAgency`, and falls back to `staticData.Stops` only when that index is completely empty. `mergeStaticAndDiscoverAgencies` skips agencies with a blank `agency_id`, so in a mixed multi-feed config (one feed declaring `agency_id`, another omitting it — legal for a single-agency feed, as the comment in `storeStaticForServer` itself notes: "or is blank — which is legal for a single-agency feed") the blank feed's stops are indexed nowhere and drop out of the union box, which previously came from `mergedbundle.Stops` and included them. That is the worst case for the fallback box: routes from a blank-`agency_id` feed get no entry in `routeAgencyIndex`, so their vehicles are unattributed and validated against precisely the union box that no longer covers their territory — every stopped vehicle in that feed counts toward `gtfs_rt_stopped_out_of_bounds_vehicles`. Seeding `seen` from (or unioning with) `staticData.Stops` would keep the old coverage while still picking up the retained collision locations.

https://github.com/OneBusAway/watchdog/blob/bd30b8d3341211f3c44b7f13a8bfcaf578661cf9/internal/gtfs/gtfs_bundles.go#L243-L263

3. `TestStoreStaticForServerFallsBackToUnionBoundingBox` never reaches the fallback it is named for. Both agencies are declared by the *same* bundle, and the merge indexes every stop under every `agency_id` that bundle declares, so `agencyBoxes["agency-B"]` is present and identical to `agencyBoxes["agency-A"]` — the `!ok` warn-and-use-union branch is not executed, and the `agencyA != agencyB` assertion holds whether or not that branch exists. The reachable fallback cases (agency-mode where the configured `AgencyID` is absent from `agency.txt`, or a declared agency whose feeds have no usable coordinates) stay untested; a second bundle declaring only `agency-B` with no stops, or a configured `AgencyID` not in `agency.txt`, would exercise it.

https://github.com/OneBusAway/watchdog/blob/bd30b8d3341211f3c44b7f13a8bfcaf578661cf9/internal/gtfs/gtfs_bundles_test.go#L984-L989

aaronbrethorst (changes requested):
This is a solid piece of work and the hard parts are right. Things I verified rather than assumed:

- The store keying is correct throughout: per-agency boxes go under `models.ServerKey(baseURL, agencyID)` and the union under the server-scoped `models.ServerKey(baseURL, "")`, which is the key the vehicle pass actually reads. That one is easy to get wrong and you got it right.
- `go vet` and the full test suite are clean on the branch.
- The design matches what #137 and #140 asked for, and keeping one shared merged bundle while storing a separate per-agency box is the right call.

Three things I'd like fixed before it lands.

**1. The rename retirement of `oba_unmatched_stop_info` is now dead code.** After this PR `fetchObaAPIMetrics` only calls `RecordLocationLastSeen`, which deliberately never retires another label set for the same stop ID, and `RecordLastSeen` has no production callers left at all. That silently undoes `438dab7` and `30fca50`: a stop that is renamed or relocated now leaves its old series live alongside the new one until the 24h sweep, double-counting in the `sum by (agency_id, stop_id) (max_over_time(...))` query we document. `TestRecordLastSeenUpdatesLabelsOnRename` still passes because it calls the dead method directly, so CI can't catch this. The two cases need to stay distinguishable: a genuine rename should still retire the old series, while a true collision keeps both.

**2. The union bounding box silently drops stops from feeds with a blank `agency_id`.** `allStops` walks only `StopsByAgency`, and `mergeStaticAndDiscoverAgencies` indexes stops only under non-empty agency IDs. The `len(stops) == 0` fallback to `staticData.Stops` catches the all-blank case but not a mix, so on a server with one normal feed and one blank-`agency_id` feed the union box loses the blank feed's coverage entirely. That's self-reinforcing: a blank-`agency_id` feed also contributes no `routeAgencyIndex` entries, so its vehicles are unattributed and then validated against the very box that no longer covers them, producing false `gtfs_rt_stopped_out_of_bounds_vehicles`. The comment right above `storeStaticForServer` calls a blank `agency_id` legal and expected, so this is reachable.

**3. `TestStoreStaticForServerFallsBackToUnionBoundingBox` doesn't test the fallback.** Both agencies are declared by the same bundle, so the merge indexes every stop under both, `agencyBoxes["agency-B"]` always exists, the `!ok` branch never runs, and `agencyA != agencyB` passes trivially because both boxes come from the same stops. I confirmed it by putting a `panic()` in the fallback branch: the test still passes. It needs an agency that genuinely has no stops of its own.

Three non-blocking notes while you're in here:

- `computeAgencyBoundingBoxes` drops `ComputeBoundingBox` errors with `if ... err == nil`, so a failure only ever surfaces indirectly as the downstream "No stops associated with agency" warning.
- The `continue` on `unionBoxErr` now skips the `observer` callback, so `gtfs_static_stops_count` and `gtfs_static_routes_count` stop being emitted for a server whose feeds yield no usable coordinates. Previously a bbox failure logged and still ran the observer.
- `docs/METRICS.md` still describes one series per `(agency_id, stop_id)`, which this PR makes untrue, and `oba_unmatched_stop_cluster_count` now counts physical locations rather than stop IDs.

Happy to re-review as soon as those three are addressed.

aaronbrethorst (changes requested):
All three blockers and all three non-blocking notes are addressed, and I verified each one in the code rather than taking the commit messages for it. This is good work.

- `RecordLastSeen` has a production caller again: `fetchObaAPIMetrics` picks it when `len(stops) == 1` and `RecordLocationLastSeen` when there's more than one, and the retirement walk in `unmatched_stop_tracker.go` deletes the old label set for a same-`stop_id` different-key entry. That's a superset of what `438dab7` and `30fca50` did, and `TestFetchObaAPIMetricsRetiresRenamedStopSeries` drives it through `fetchObaAPIMetrics` rather than calling the method directly, so CI can actually catch a regression now. That was the part that mattered.
- The blank-`agency_id` hole is closed properly. Bucketing those feeds under a `""` sentinel that `allStops` walks and `computeAgencyBoundingBoxes` skips is a cleaner fix than appending `staticData.Stops`, and the test pins both halves — the union covers the blank feed, agency-A's own box doesn't.
- `TestStoreStaticForServerFallsBackToUnionBoundingBox` tests the fallback now. `bundleB` is built with `nil` stops, so `StopsByAgency["agency-B"]` is never created, the `!ok` branch genuinely runs, and the assertion is against a union value neither A's nor C's own box could produce. That's exactly the shape I was asking for.
- `computeAgencyBoundingBoxes` returns its errors, the `unionBoxErr` `continue` is gone so the observer always runs, and both are covered by tests.

I also checked the keying invariants specifically, since this PR is squarely in that territory: per-agency boxes under `ServerKey(baseURL, agencyID)` with the union under the server-scoped key, `routeAgencyIndex.Set` still on the raw base URL, both `tracked(...)` wrappers intact with label sets unchanged. All correct.

One thing to fix before it lands, and it's one line.

`docs/METRICS.md` line 115 still says:

> **The bounding box is still server-wide:** `gtfs_rt_stopped_out_of_bounds_vehicles` is attributed per agency, but the box it tests against is computed over the union of every configured static feed's stops... treat this metric as a loose bound rather than a precise one.

That's the thing this PR fixes, documented as not having been fixed. You corrected the other stale rows in this file, so this one just got missed. It matters more than a normal doc nit because it actively tells operators not to tighten alerts on the gauge you just made precise — the improvement ships and nobody uses it. The sentence is still true for unattributed vehicles and for an agency that fell back to the union, so it needs rewording rather than deleting.

Non-blocking, for your judgment:

- `CLAUDE.md` says `geo.BoundingBoxStore` publishes "the same box" under the server-scoped key, which this PR makes untrue. `03bfdea` and `ecf8845` updated architecture docs alongside changes like this, so there's precedent for doing it here.
- The per-agency index appends a full `Stop` copy under every `agency_id` the bundle declares. For a single pre-merged multi-agency bundle — the configuration the NOTE you deleted actually recommended — that's an N-times blow-up of the stop set, and every agency box ends up equal to the union anyway, so it costs memory and buys nothing in that shape. The `Memory cost stays O(bundles) not O(bundles × agencies)` comment just below now describes only the `StaticData` pointer, not the stop data it carries. Worth a comment saying so.
- The comment above the `len(stops) > 1` branch says duplicate locations "came from different scrapes (we fetch static data daily)". They didn't — they're colliding feeds within one merged snapshot. Your own new METRICS.md bullet says the right thing. Since that comment is what justifies suppressing rename retirement, I'd rather it name the real cause.
- The new bbox error branches log but don't report to Sentry, while `storeStaticForServer` does dual-report for the collision cases right next to them. The old path only logged too, so this is an existing gap widening rather than a new one.
- `getStopLocationsByIDs` has no per-ID fall-through once `StopsByAgency[agencyID]` exists, so a stop that lives only in another agency's feed now increments `oba_unmatched_stop_unresolved`, which METRICS.md documents as meaning something else entirely. Arguably more correct, but it's a silent change to an alerting metric.

Separately, the CLA is blocking this and your four other open PRs. One of two committers has signed; the unsigned commits are authored as `Mohamed Ahmed Aboomar <aboomar@Mohameds-MacBook-Air.local>`, a local machine hostname rather than a real address, so the bot's advice to add that email to your GitHub account won't work. Rewrite the authorship and force-push:

```
git push --force-with-lease
```

Fix the METRICS.md line and this is ready to go.

0xaboomar:
> The comment above the len(stops) > 1 branch says duplicate locations "came from different scrapes (we fetch static data daily)". They didn't — they're colliding feeds within one merged snapshot. Your own new METRICS.md bullet says the right thing. Since that comment is what justifies suppressing rename retirement, I'd rather it name the real cause.

To clarify what I meant by that comment: I wasn’t referring to collisions between feeds within the same merged snapshot. We don’t allow that case.

The scenario I was referring to happens over time at the Prometheus level. A `stop_id` may be reported with one set of coordinates, then after a later static bundle refresh, the same `stop_id` may have corrected or relocated coordinates. Watchdog replaces the old label set with the new one via `RecordLastSeen`, while Prometheus may still retain samples from the old series as history.

So the “different scrapes” comment was referring to this historical Prometheus behavior, not a collision in Watchdog’s current data model.

This is just for clarification, though the `len(stops) > 1` path is no longer used.

0xaboomar:
> The per-agency index appends a full Stop copy under every agency_id the bundle declares. For a single pre-merged multi-agency bundle — the configuration the NOTE you deleted actually recommended — that's an N-times blow-up of the stop set, and every agency box ends up equal to the union anyway, so it costs memory and buys nothing in that shape. The Memory cost stays O(bundles) not O(bundles × agencies) comment just below now describes only the StaticData pointer, not the stop data it carries. Worth a comment saying so.

Hey @aaronbrethorst, Following up on this, I made a design change to how the per-agency and union bounding boxes are calculated.

Previously, we maintained a per-agency stop index, which meant copying the full set of stops for every agency declared by a bundle. For a pre-merged multi-agency bundle, this added significant memory overhead while providing little value, since each agency in that bundle would receive the same set of stops.

The new approach calculates the bounding boxes directly from the original source bundles. Each agency keeps only the four values needed to build its bounding box (`minLat`, `maxLat`, `minLon`, `maxLon`), while a separate accumulator builds the server-wide union. This preserves the per-agency bounding boxes without keeping per-agency copies of the stops.

For example, with a 5,000-stop bundle containing 10 agencies, the old approach could store 50,000 `Stop` entries in the per-agency index. The new approach processes the same stops while retaining only 44 floats for bounding-box state: 4 for each of the 10 agencies plus 4 for the union.

This also lets us keep a single shared server-wide stop set for stop resolution, while the agency-specific data is limited to the bounding boxes where that distinction is actually needed.

aaronbrethorst:
### Code review

Found 1 issue:

1. The `oba_unmatched_stop_info` table row still carries "One stop ID can produce multiple series", which is a leftover from the multi-location design this branch later reverted in `refactor(metrics): use singular stop map for runtime lookup and metrics`. As shipped, `getStopLocationsByIDs` returns at most one stop per ID (first occurrence from the merge) and `RecordLastSeen` walks the tracker deleting every other `stopKey` with the same `StopID` before inserting the new one — so within an `(agency_id, stop_id)` there is always exactly one live series. The retention bullet ten lines below says exactly that ("Each `(agency_id, stop_id)` resolves to one stop in Watchdog's merged static bundle... the old label set is retired immediately"), so the two rows contradict each other, and an operator writing `sum by (agency_id, stop_id) (max_over_time(...))` is told to expect a multiplicity that cannot occur. The description above it ("for unmatched physical stop locations") reads the same way. `docs: reflect singular stop index...` caught the other rows in this file; this one got missed.

https://github.com/OneBusAway/watchdog/blob/395ec46078042fbf90a3d08aa4308f2c7ccaa854/docs/METRICS.md#L135-L137

aaronbrethorst (approved):
The blocker from last round is fixed. `docs/METRICS.md` line 115 now reads "Bounding box accuracy depends on attribution" and correctly keeps the loose-bound caveat for unattributed vehicles rather than deleting it, which is what I was asking for.

You also went further than I asked on the non-blocking notes, and the result is better for it. Removing the `StopsByAgency` index entirely and computing boxes from the source feeds with four extrema per agency is a much better answer than the comment-the-blow-up suggestion I made — it makes the memory question disappear instead of documenting it. The bbox error paths now dual-log and report to Sentry with `server_name` + `agency_id`, and the CLAUDE.md `BoundingBoxStore` bullet is accurate again.

I checked the keying invariants specifically, since that's where this PR lives: per-agency boxes under `ServerKey(baseURL, agencyID)`, union under the server-scoped key, `routeAgencyIndex.Set` still on the raw base URL, `tracked(...)` wrappers and label sets unchanged, and nothing keyed on `agency_id` alone. All correct. CI is green and coverage is up 1.7%.

So this is approved on the merits. One thing has to happen before it can land, and it isn't your fault.

## Please rebase — #144 landed underneath you

I merged #144 a few minutes ago. It added a `ctx context.Context` first parameter to `fetchObaAPIMetrics`, and this branch rewrites the same file. Git merges the two cleanly and `go build ./...` passes, so GitHub reports this PR mergeable and your CI run is green — but that run was against a base without #144. I trial-merged this branch onto current `main` and the test build fails:

```
internal/metrics/oba_rest_api_metrics_test.go:153:169: not enough arguments in call to fetchObaAPIMetrics
internal/metrics/oba_rest_api_metrics_test.go:160:169: not enough arguments in call to fetchObaAPIMetrics
internal/metrics/oba_rest_api_metrics_test.go:187:167: not enough arguments in call to fetchObaAPIMetrics
```

Every other package passes; it's only those three call sites, and they just need the new leading `ctx` argument. Merge `main` in, add it, and push — no re-review needed unless the resolution changes behavior, and I'll merge it as soon as it's green.

## One follow-up, not a blocker

`docs/METRICS.md` line 136 still says "One stop ID can produce multiple series" in the `oba_unmatched_stop_info` row. That's a leftover from the multi-location design this branch reverted, and it contradicts your own retention bullet ten lines below, which correctly says each `(agency_id, stop_id)` resolves to one stop. I confirmed the code agrees with line 146: `RecordLastSeen` is the only caller now, and it deletes the series for every same-`StopID` key that differs, so exactly one series survives per stop ID.

I'm not holding the merge on it — unlike the line 115 problem, this one doesn't cost anyone anything in practice, since a defensive `sum by (agency_id, stop_id)` returns the same answer at cardinality 1. Fold it into the rebase if it's easy.

0xaboomar:
Hey @aaronbrethorst, I've rebased the branch, resolved the conflict, and applied the mentioned non-blocking changes; This PR is ready to be merged.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
The rebase is clean and my approval stands. I verified that rather than assuming it.

I reconstructed your patch against both its old base and current `main` and diffed the two: 15 files, +661/-141 either way, and the only differing lines in the whole patch set are the three `context.Background()` arguments and the one `docs/METRICS.md` cell. Nothing else in your contribution moved, and nothing from #144 was dropped on the way through — the forward and reverse diffs against `main` are exact mirrors. `go vet` and the full suite pass for me on the rebased head, and again on a trial merge with #130. The three `fetchObaAPIMetrics` call sites that failed my trial merge last round are fixed.

The `oba_unmatched_stop_info` follow-up is folded in too — the row now agrees with the retention bullet below it.

Merging. Thanks for turning this around quickly, and for going past what I asked on the earlier rounds — computing the boxes from the source feeds instead of retaining a per-agency stop index is a genuinely better answer than the one I suggested.

One thing I want to flag rather than sit on, since I found it while verifying and it is **not** something the rebase introduced — it was in the version I approved, and I missed it then.

## Agency-mode with a blank-`agency_id` sibling feed gets a narrower box than it did on `main`

`computeBoundingBoxes` only accumulates a stop into `byAgency[id]` for agencies that the *containing bundle* declares, and it skips blank `agency_id` rows. In agency-mode `storageAgencies` is forced to the configured `server.AgencyID`, so if one configured feed declares that ID and a sibling feed omits `agency_id`, `agencyBoxes[server.AgencyID]` exists and the sibling's stops never reach the stored box. On `main` the stored box was `ComputeBoundingBox(merged.Stops)` — every feed.

I confirmed it with a throwaway test: two feeds, one declaring `agency-X` with stops around (10,10)–(11,11), one with a blank `agency_id` around (50,50)–(51,51).

```
stored agency-mode box: {MinLat:10 MaxLat:11 MinLon:10 MaxLon:11}
box main would store:   {MinLat:10 MaxLat:51 MinLon:10 MaxLon:51}
blank-feed stop (50,50) inside stored box? false
```

So every stopped vehicle in that sibling feed's territory counts toward `gtfs_rt_stopped_out_of_bounds_vehicles`. It's the same blank-`agency_id` class I raised in the first round — you fixed it for the union box in `62f6eef`, and this is the agency-box half of it.

It needs agency-mode *plus* multiple static feeds *plus* one of them omitting `agency_id`, which is narrow enough that I'd rather land the improvement now than hold it for another round. Filed as a follow-up with the repro; the fix is probably to fold blank-`agency_id` feeds into the configured agency's accumulator in agency-mode, the way the union already does.

PR #142, opened 2026-08-29
feat(prometheus): add operational alerting rules

0xaboomar:
## Summary by CodeRabbit

- **Documentation**
  - Updated bounding-box validation guidance for `gtfs_rt_stopped_out_of_bounds_vehicles`.
  - Clarified that vehicles associated with an agency are checked against that agency’s bounding box.
  - Vehicles without a resolved agency continue to use the server-wide union bounding box.
  - Documented that the empty-`agency_id` series represents a looser bound than per-agency series.

### DIFF
--- a/prometheus/rules/operational.yml
+++ b/prometheus/rules/operational.yml
@@ -1,0 +1,23 @@
+groups:
+  - name: watchdog-operational
+    rules:
+      - alert: ObaAPIDown
+        expr: oba_api_status == 0
+        for: 2m
+        labels:
+          severity: critical
+        annotations:
+          summary: "OneBusAway API unavailable for {{ $labels.server_name }}"
+          description: >-
+            The OneBusAway API at {{ $labels.server_url }} has reported
+            unavailable for 2m+ (oba_api_status={{ $value }}).
+      - alert: GTFSBundleExpiringSoon
+        expr: gtfs_bundle_days_until_earliest_expiration < 3
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: "GTFS bundle expiring soon for {{ $labels.agency_name }} ({{ $labels.agency_id }})"
+          description: >-
+            The earliest GTFS bundle for {{ $labels.agency_name }} ({{ $labels.agency_id }})
+            at {{ $labels.server_url }} expires in {{ $value }} days, below the 3-day threshold.


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
The alert rules are the good part of this PR and I want to keep them. I verified both expressions against the code rather than just reading them:

- `oba_api_status` is declared with exactly `{server_name, server_url}`, and your alert references only those two.
- `gtfs_bundle_days_until_earliest_expiration` carries `{agency_id, agency_name, server_name, server_url}`, and your annotations use `agency_name`, `agency_id`, and `server_url`.
- Both land under the existing `rule_files: /etc/prometheus/rules/*.yml` glob and the docker-compose mount, and neither alert name collides with the three rule files already there. They also match the example expressions already recommended in `docs/METRICS.md`.

So the Prometheus half is ready. Three things to fix first.

**Blocker: the `downloadGTFSBundle` context change makes `TestRefreshGTFSBundlesReadsLiveConfig` fail most of the time.** Once the request carries ctx, the test's deferred `cancel()` aborts an in-flight download before the server is drained, and the handler's `w.Write(bundle)` fails into `t.Errorf`. Measured: 0 failures in 30 runs on `main`, 21 in 30 on this branch. CI passes only because Linux runners buffer the entire 1.8 MB response. The fix is to drain before cancelling — `ts.Close()` before `cancel()` in the deferred cleanup, or join the goroutine with a `done` channel.

Worth knowing: this change, both the `gtfs_bundles.go` hunk and the test, is byte-identical in #144. Whichever you fix first, the other's copy should be dropped so the two don't diverge.

**The title doesn't describe the diff.** It says "scope per-agency vehicle metrics", but that work landed in #133 and none of it is here. This PR is request cancellation for bundle downloads, a test cleanup change, an environment label, and two alert rules. Since we merge with merge commits, the wrong title goes into main's history right next to the PR that actually did that work. Please retitle.

**`environment: production` on the local stack.** `prometheus/prometheus.yml` is only mounted by `docker-compose.yml`, which is how we run the whole stack locally. This stamps `environment="production"` on every series on a developer's laptop, and nothing consumes the label today — `alertmanager.yml` groups on `["alertname", "severity"]`, and no rule or dashboard reads it. If the goal is labelling a real deployment, that belongs wherever the production Prometheus config lives, not here.

Fix the test and retitle, and this is good to go. The alerts themselves I'd merge as-is.

I tested two candidate fixes on this branch so you don't have to guess:

- Just reordering the cleanup to `ts.Close()` before `cancel()` helps a lot but isn't sufficient: 21/30 failures drops to 3/30.
- Marking the shutdown explicitly and not asserting on writes that fail because of it is clean: **0 failures in 60 runs, and clean under `-race`.**

That second one is this:

```diff
 	var (
+		shuttingDown  atomic.Bool
 		firstTickOnce sync.Once
 		addedOnce     sync.Once
 		requests      sync.WaitGroup
@@
-		if _, err := w.Write(bundle); err != nil {
+		if _, err := w.Write(bundle); err != nil && !shuttingDown.Load() {
 			t.Errorf("write GTFS fixture: %v", err)
 		}
@@
 	defer func() {
+		shuttingDown.Store(true)
 		cancel()
 		requests.Wait()
 		ts.Close()
 		http.DefaultClient.CloseIdleConnections()
 	}()
```

plus `"sync/atomic"` in the imports. The write assertion still guards the real case it was added for in `ecf8845` — a mid-test write failure — and only goes quiet once the test has deliberately begun tearing down. Take it or something equivalent, whichever you prefer.

aaronbrethorst:
Correcting my earlier automated review comment on this PR: it said "No issues found", and that was wrong.

The review did flag the test-cancellation interaction internally, but scored it below its posting threshold on the assumption it might only be theoretical. It isn't. Measured on this branch:

- `main`: 0 failures in 30 runs
- this branch: 21 failures in 30 runs

Details are in the formal review I've just submitted. Sorry for the noise.

aaronbrethorst (changes requested):
The blocker is fixed. You applied the guard essentially as written, and the teardown now stores the shutdown flag, cancels, drains, then closes, so the write assertion still catches a genuine mid-test failure and only goes quiet once teardown has started. That's what I was after.

I re-validated the alert rules against the code rather than just re-reading them, and they're still clean:

- `oba_api_status` is declared with exactly `{server_name, server_url}` and your annotations use only those.
- `gtfs_bundle_days_until_earliest_expiration` carries `{agency_id, agency_name, server_name, server_url}` and your annotations use a subset.
- Both are gauges compared against scalars, so no `rate()`/`increase()` misuse.
- The group name and both alert names don't collide with the three existing rule files, and the file lands under the `rule_files` glob and the compose mount.
- `for: 2m` is about four collection cycles at the default 30s fetch interval, so it won't flap. `checkBundleExpiration` returns early on every error path and never zeroes the gauge, so a failed fetch can't spuriously trip the `< 3` threshold.

Dropping the `environment: production` label was the right call.

Two things left, neither of them about the alerts.

**The title still describes #133's work.** It says "scope per-agency vehicle metrics", and none of that is in this diff — it landed in #133. What's actually here is a one-line request-context change, its test fallout, and two alert rules. We merge with merge commits, so this title goes into `main`'s history right next to the PR that did the thing it names. Please retitle to something like "add Prometheus operational alert rules".

**This and #144 have now diverged on the same code.** Both change `downloadGTFSBundle` to `http.NewRequestWithContext` and both rewrite the same `var` block and `defer` in `refresh_live_config_test.go`, but differently: you drain with a `requests sync.WaitGroup` here, and join the refresh goroutine with a `done` channel in #144. Those hunks overlap textually, so whichever merges second will conflict, and we'd end up with two different rationales for one teardown.

#144's version is the one to keep. It's the goroutine join I suggested, and it drops the `WaitGroup` I asked you to delete — `httptest.Server.Close()` already blocks until outstanding handlers return, so `requests.Wait()` isn't buying anything, and `Add(1)` in the handler while `Wait()` has a waiter registered is the documented misuse even though the window is tiny.

So: drop `internal/gtfs/gtfs_bundles.go` and `internal/gtfs/refresh_live_config_test.go` from this PR entirely and let #144 carry them. This becomes an alerts-only PR of one new file, which is a much easier thing to merge, and the conflict disappears.

Separately, the CLA is blocking this and your four other open PRs. One of two committers has signed; the unsigned commits are authored as `Mohamed Ahmed Aboomar <aboomar@Mohameds-MacBook-Air.local>`, a local machine hostname rather than a real address, so the bot's suggestion to add that email to your GitHub account won't work. Rewrite the authorship and force-push:

```
git push --force-with-lease
```

Retitle, drop the two `internal/gtfs` files, and the alerts are good to merge as-is.

0xaboomar:
Hey @aaronbrethorst, I’ve reverted the changes in `internal/gtfs/gtfs_bundles.go` and `internal/gtfs/refresh_live_config_test.go`, as suggested, since they’ll be handled in #144. I’ve also updated the PR title to reflect the changes included in this PR.

This PR is ready to be merged now.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both asks from last time are done: the title now describes the diff, and dropping `internal/gtfs/gtfs_bundles.go` and `refresh_live_config_test.go` leaves this as an alerts-only PR of one new file, which removes the textual conflict with #144. Thanks for taking the deletion route rather than patching around it — that was the cleaner of the two options.

I re-validated the rules against the code a third time rather than trusting my earlier passes:

- `oba_api_status` is declared with exactly `{server_name, server_url}` and the annotations use only those two. `server_ping.go` writes 0 on failure and 1 on success, so `== 0` is the right comparison and reachable in both directions.
- `gtfs_bundle_days_until_earliest_expiration` carries `{agency_id, agency_name, server_name, server_url}`; the annotations use a subset.
- Both are gauges compared against scalars, so there's no counter-without-`rate()` problem.
- `watchdog-operational` and both alert names don't collide with `feed-staleness`, `watchdog-meta`, or `prediction-quality`, and the file lands under the `rule_files` glob and the compose mount.
- `checkBundleExpiration` returns early on every error path and never zeroes the gauge, so a failed fetch can't spuriously trip `< 3`.

One thing I'm fixing on the way in rather than sending back: the PR body is still a CodeRabbit summary of #141's bounding-box work and describes nothing in this diff. Since we merge with merge commits, that would land in `main`'s history. I'll override the merge commit body — no action needed from you, but worth clearing stale auto-summaries before a merge in future.

Merging.

PR #144, opened 2026-08-30
fix: context cancellation graceful shutdown

### LINKED ISSUE #143: Propagate context cancellation through background refresh and metrics requests
Several HTTP requests and background loops ignore cancellation contexts, allowing in-flight requests or refresh goroutines to continue during shutdown and test teardown. This can cause connection errors, goroutine leaks, and flaky CI failures.

0xaboomar:
closes #143 

## Summary by CodeRabbit

* **New Features**
  * Added graceful shutdown handling for interrupt and termination signals.
  * In-progress network operations now stop promptly when the application is shutting down or canceled.
  * Configuration refreshes and data collection honor cancellation and shutdown requests.

* **Bug Fixes**
  * Normal server shutdowns are no longer reported as fatal errors.
  * Unexpected serving or shutdown failures are reported more accurately.

### DIFF
--- a/cmd/watchdog/main.go
+++ b/cmd/watchdog/main.go
@@ -2,11 +2,14 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"log/slog"
 	"net/http"
 	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/getsentry/sentry-go"
@@ -82,8 +85,8 @@ func main() {
 	// This context will be used to manage the application's lifecycle and cancel operations when needed.
 	// It allows us to gracefully shut down the application and clean up resources.
 	// we will use it to cancel and clean up routines when the application is shutting down.
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
 
 	// Create a new HTTP client with a connection pool
 	// This client will be reused across the application to avoid creating new connections for each request.
@@ -186,9 +189,28 @@ func main() {
 	}
 
 	logger.Info("starting server", "addr", srv.Addr, "env", cfg.Env)
-	err = srv.ListenAndServe()
-	report.ReportError(err, sentry.LevelFatal)
-	report.FlushSentry()
-	logger.Error(err.Error())
-	os.Exit(1)
+
+	serverErr := make(chan error, 1)
+	go func() {
+		serverErr <- srv.ListenAndServe()
+	}()
+
+	select {
+	case <-ctx.Done():
+		logger.Info("Shutting down HTTP server")
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if err := srv.Shutdown(shutdownCtx); err != nil {
+			report.ReportError(err, sentry.LevelError)
+			logger.Error("HTTP server shutdown failed", "error", err)
+		}
+	case err := <-serverErr:
+		if err == nil || errors.Is(err, http.ErrServerClosed) {
+			return
+		}
+		report.ReportError(err, sentry.LevelFatal)
+		report.FlushSentry()
+		logger.Error(err.Error())
+		os.Exit(1)
+	}
 }
--- a/internal/app/metrics_collector.go
+++ b/internal/app/metrics_collector.go
@@ -74,7 +74,7 @@ func (app *Application) StartMetricsCollection(ctx context.Context) {
 func (app *Application) collectForScope(ctx context.Context, server models.ObaServer, scope config.Scope) {
 	switch s := scope.(type) {
 	case config.AgencyScope:
-		app.CollectMetricsForServer(server)
+		app.CollectMetricsForServer(ctx, server)
 	case config.ServerScope:
 		app.collectForServerScope(ctx, server, s)
 	default:
@@ -112,7 +112,7 @@ func (app *Application) collectForServerScope(ctx context.Context, server models
 	// Server-ping once per server (the /current-time.json endpoint takes no
 	// agency parameter; per the metrics relabeling in this redesign, the
 	// ObaApiStatus gauge is server-scoped, not agency-scoped).
-	ok := app.MetricsService.ServerPing(server)
+	ok := app.MetricsService.ServerPing(ctx, server)
 	if !ok {
 		app.ConfigService.BackoffStore.UpdateBackoff(server.ServerKey())
 		app.Logger.Info("Skipping server-scope collection due to ping failure",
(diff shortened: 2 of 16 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `TestRefreshGTFSBundlesReadsLiveConfig` becomes flaky (~10% locally). `downloadGTFSBundles` fans out one goroutine per server, so `/initial.zip` and `/added.zip` (1.8 MB each) are served concurrently. The test closes `sawAdded` from the `/added.zip` handler and then immediately runs `cancel()` in the deferred cleanup — but now that `downloadGTFSBundle` builds its request with `http.NewRequestWithContext(ctx, ...)`, that `cancel()` aborts the still-in-flight `/initial.zip` request, its handler's `w.Write(bundle)` returns an error, and the `t.Errorf("write GTFS fixture: %v", err)` added in ecf8845 fails the test. Before this PR the request carried no context, so `cancel()` could not interrupt a download and the write always completed. A minimal reproduction of the same shape (two concurrent 1.8 MB responses, cancel right after the second handler's write) reports the write error 3/30 runs with a request context and 0/60 without. Moving `cancel()` after `ts.Close()`, or joining the `refreshGTFSBundles` goroutine before cancelling, would close the window.

https://github.com/OneBusAway/watchdog/blob/1180c0044e7e70e6f67730c3549f9e3f39a956a8/internal/gtfs/refresh_live_config_test.go#L78-L90

aaronbrethorst (changes requested):
Thanks for this — context plumbing is the right change and most of it is carefully done. The things I checked that hold up:

- Every background goroutine still exits cleanly on `ctx.Done()`, and you didn't reintroduce a blocking sleep anywhere. `BackoffStore` is still the backoff mechanism, and the `time.Sleep(interval)` to cancellable-timer swap in `refreshConfig` is exactly right.
- `serverErr` is buffered, so the `ListenAndServe` goroutine can't leak if the shutdown path wins the select.
- `signal.NotifyContext` plus `srv.Shutdown` with a bounded timeout is the correct shape, and the `defer report.FlushSentry()` already at the top of `main` covers the graceful path, so Sentry events aren't dropped on the way out.

The blocker is `TestRefreshGTFSBundlesReadsLiveConfig`. Adding the request context makes `cancel()` able to abort an in-flight download, and the test's deferred cleanup calls `cancel()` before the server is drained. `refreshGTFSBundles` fans out one goroutine per server on a 10ms tick with a 1.8 MB fixture, so when the test body returns there is almost always a request mid-write; the handler's `w.Write(bundle)` then fails and trips the `t.Errorf` added in `ecf8845`.

Measured:

- `main`: 0 failures in 30 runs
- this branch: 21 failures in 30 runs

CI is green because GitHub's Linux runners buffer the whole 1.8 MB response, so the handler never notices the client is gone. It reproduces reliably on macOS.

The fix is one line: drain before cancelling, so the deferred cleanup does `ts.Close()` and then `cancel()`. Better still, join the `refreshGTFSBundles` goroutine with the `done` channel pattern you already applied to the three sibling tests in this same PR.

Two smaller things, neither blocking:

- The `requests sync.WaitGroup` calls `Add(1)` inside the handler while `Wait()` runs in cleanup. `httptest.Server.Close()` already blocks until outstanding requests finish, so once the ordering above is fixed the WaitGroup can simply be deleted.
- Now that ctx reaches `ServerPing`, `FetchObaAPIMetrics`, and the RT fetch, a SIGTERM landing mid-tick will report several `context canceled` errors to Sentry at `LevelError`. A `ctx.Err() != nil` guard before reporting would keep shutdown quiet. Fine as a follow-up.

Fix the test ordering and I'll re-review promptly.

I tested two candidate fixes on this branch so you don't have to guess:

- Just reordering the cleanup to `ts.Close()` before `cancel()` helps a lot but isn't sufficient: 21/30 failures drops to 3/30.
- Marking the shutdown explicitly and not asserting on writes that fail because of it is clean: **0 failures in 60 runs, and clean under `-race`.**

That second one is this:

```diff
 	var (
+		shuttingDown  atomic.Bool
 		firstTickOnce sync.Once
 		addedOnce     sync.Once
 		requests      sync.WaitGroup
@@
-		if _, err := w.Write(bundle); err != nil {
+		if _, err := w.Write(bundle); err != nil && !shuttingDown.Load() {
 			t.Errorf("write GTFS fixture: %v", err)
 		}
@@
 	defer func() {
+		shuttingDown.Store(true)
 		cancel()
 		requests.Wait()
 		ts.Close()
 		http.DefaultClient.CloseIdleConnections()
 	}()
```

plus `"sync/atomic"` in the imports. The write assertion still guards the real case it was added for in `ecf8845` — a mid-test write failure — and only goes quiet once the test has deliberately begun tearing down. Take it or something equivalent, whichever you prefer.

aaronbrethorst (approved):
Approving this. The blocker is fixed more thoroughly than I asked for, and I re-checked the concurrency surface rather than assuming.

What I verified at `97163d7`:

- `TestRefreshGTFSBundlesReadsLiveConfig` now does both halves: the `shuttingDown` guard on the write assertion *and* the `done`-channel join of the refresh goroutine, so the teardown drains before `ts.Close()`. That closes the `t.Errorf`-after-test-completion hazard too, not just the flake.
- The `requests sync.WaitGroup` is gone, as suggested. Good, that was the redundant part.
- The shutdown path is right: `signal.NotifyContext` installed before any goroutine starts, `serverErr` buffered so the `ListenAndServe` goroutine can't leak when the cancel branch wins, a bounded 10s `srv.Shutdown`, `ErrServerClosed` treated as a clean exit, and a real bind failure still fatal with an explicit `FlushSentry()` before `os.Exit(1)` since the defers won't run there.
- Context propagation is now complete: the `context.Background()` seams in `server_ping.go` and `vehicle_metrics.go` are gone, and there's no `context.Background()`/`context.TODO()` left in non-test code under `internal/` or `cmd/`.
- Backoff is untouched. `BackoffStore` is still the mechanism, `refreshConfig`'s sleep became a cancellable timer, and no blocking retry crept into the collection path.

Two notes, neither blocking:

- The `ctx.Err() != nil` guard to keep shutdown quiet in Sentry is still outstanding. That was mine to file as a follow-up and I'm not holding the PR on it.
- CodeRabbit's suggestion to return after a failed server-scope RT fetch should not be taken. It contradicts CLAUDE.md and the comment directly above that branch, and it was raised and rejected on #133 for the same reason. Leaving it as-is is correct.

One thing stands between this and merge, and it isn't the code: the CLA check. One of two committers has signed. The unsigned commits are authored as `Mohamed Ahmed Aboomar <aboomar@Mohameds-MacBook-Air.local>`, which is a local machine hostname rather than a real address, so the bot's suggested fix (adding the email to your GitHub account) won't work here. You'll need to rewrite the authorship to the address on your GitHub account and force-push:

```
git push --force-with-lease
```

Same applies to #130, #136, #141, and #142 — all five branches have commits from that address.

Sequencing note: this PR and #142 both change `downloadGTFSBundle` to use `http.NewRequestWithContext` and both rewrite the same teardown in `refresh_live_config_test.go`, but differently — you use the `done` channel here and a `sync.WaitGroup` there. Whichever lands second will conflict. Yours here is the better version, so I'd keep it and drop both `internal/gtfs` hunks from #142, leaving that PR to carry only the alert rules.

aaronbrethorst:
### Code review

aaronbrethorst:
Re-verified after the force-push before merging. The rebase was an authorship rewrite for the CLA and nothing else — `97163d7` and the current head `dd367ddf` have the identical tree `224e00ae`, so the content is byte-for-byte what I approved.

I re-read the shutdown block rather than relying on the earlier pass: `serverErr` is buffered at 1 so the `ListenAndServe` goroutine can't leak when the `ctx.Done()` branch wins, `srv.Shutdown` is bounded at 10s, `http.ErrServerClosed` is treated as a clean exit, and `defer report.FlushSentry()` at line 106 covers the graceful path so the shutdown-failure report isn't lost.

Merging.

PR #145, opened 2026-09-05
fix/reuse-metrics-response

### LINKED ISSUE #139: Server-mode performs redundant `/metrics.json` requests
### Problem

In server-mode, `collectForServerScope` (`internal/app/metrics_collector.go:204`) calls `FetchObaAPIMetrics` once for every live agency.

However, all of these calls target the same OBA `/api/where/metrics.json` endpoint.

`probeLiveAgencies` has already fetched this exact response at the beginning of the collection cycle, meaning that with `N` live agencies we can currently make **N redundant HTTP requests per tick**.

There is already a `TODO(server-mode dedup)` at `metrics_collector.go:204`, with a related reference in `oba_rest_api_metrics.go:55`.

### Proposed fix

Have `probeLiveAgencies` return the parsed `/metrics.json` response and pass it into `FetchObaAPIMetrics` so the response can be reused throughout the same collection cycle.

A likely implementation is to make the additional argument optional on `MetricsService.FetchObaAPIMetrics`, allowing existing agency-mode callers to remain unchanged.

This requires parameter plumbing through the relevant layers and a small parsed-response cache scoped to a single metrics tick.

### Priority

This is **not currently urgent**. The `/metrics.json` endpoint returns a small JSON payload and is designed for high-frequency polling, so at a 30-second collection interval the current overhead is limited.

We should prioritize this if:

- OBA begins rate-limiting `/metrics.json`.
    
- The number of agencies per server grows significantly.
    
- Metrics collection latency becomes noticeable.

ahmadmuhamadi:
Fixes #139 

## Problem

In server mode, `probeLiveAgencies` fetches `/api/where/metrics.json` at the
top of the tick and keeps only `entry.AgencyIDs`. The per-agency loop then
reaches `FetchObaAPIMetrics`, which fetches the same URL again for each
agency. The endpoint takes no agency parameter, so all N+1 responses are
identical.

## Change

`probeLiveAgencies` now returns the parsed response asw . It's held in
`collectForServerScope` and threaded through `collectAgencyChecks` into
`FetchObaAPIMetrics`, where a new `prefetch *OBAMetrics` parameter skips the
fetch when non-nil. Agency mode passes `nil` and is unchanged.

Emission logic is untouched. Removes the `TODO(server-mode dedup)` and the
matching doc-comment paragraph.

**Result: `/metrics.json` requests per server-mode tick go from N+1 to 1.**

## Tests

`TestServerScopeFetchesMetricsOncePerTick` runs one server-mode tick over two
live agencies and asserts one request reaches the endpoint, and that both
agencies emit their own values from that response. The stub returns an empty
`agencyIDs` after the first call, so any refetch would trip the guard and emit
nothing. Verified failing on the old behaviour.

## Summary by CodeRabbit

- **Performance**
  - Server-wide metrics collection now fetches shared metrics data once per collection interval, reducing repeated endpoint requests.
  - Agency-specific collection continues to fetch metrics independently.

- **Reliability**
  - Agency metrics are populated consistently from the shared response during server-scope collection.
  - Metrics now correctly reflect relocated and merged stops, including retiring outdated location data.

- **Tests**
  - Added coverage confirming single-fetch collection, agency metric recording, and correct stop-location handling.


0xaboomar (changes requested):
Hey @ahmadmuhamadi, The prefetch optimization looks good, great work! 

There are just a few leftover comments and docs from the old behavior that should be cleaned up:

1. `internal/app/metrics_collector.go:226-246`: Remove the entire `TODO(server-mode dedup)` block. It describes the old N+1 fetch problem and proposes the same solution that is now implemented. The existing loop comment around lines 218-225 is still accurate and can stay.

2. `internal/metrics/oba_rest_api_metrics.go:53-56`: Update the stale N+1 documentation. It currently says that server mode performs the same `/api/where/metrics.json` request once per live agency and points to the TODO above. This is no longer true. Something like this would reflect the new behavior:

   ```go
   // In server-mode this function is called once per live agency per tick,
   // but reuses the response prefetched by probeLiveAgencies instead of
   // issuing another HTTP request. Agency-mode callers pass nil and fetch
   // their own response.
   ```

3. `internal/metrics/oba_rest_api_metrics.go:40-70`: The function documentation also needs to be updated. It currently says the function always performs an HTTP GET and that `client` must be non-nil, which is no longer always true. It should explain that when `prefetch` is nil, the function fetches and decodes the endpoint itself; when `prefetch` is provided, it uses that response and skips the HTTP request. The `prefetch` parameter should also be documented, and `client` should be described as required only when `prefetch == nil`.

4. `internal/app/metrics_collector.go:135`: Small whitespace typo:

   ```go
   // Treat as "no agencies live" for this tick; static bundles st    ay
   ```

   should be:

   ```go
   // Treat as "no agencies live" for this tick; static bundles stay
   ```

5. `internal/app/metrics_collector_test.go:126-127`: The test comment is incomplete and still describes the old N+1 behavior. Update it to describe what the test now verifies:

   ```go
   // Server mode should fetch /api/where/metrics.json once per tick.
   // The parsed response is reused by each live agency's
   // FetchObaAPIMetrics call.
   ```

6. `internal/app/metrics_collector.go:270-271`: This comment reads more like a commit note than a code comment:

   ```go
   // nil prefetch: agency-mode has no /metrics.json probe ahead of this call,
   // so FetchObaAPIMetrics fetches the endpoint itself. Behavior unchanged.
   ```

   I would either remove it or shorten it to:

   ```go
   // nil: agency-mode fetches its own response.
   ```

One optional improvement: `prefetch` works, but something like `prefetchedMetrics` or `cachedMetrics` would make its purpose clearer. If you decide to rename it, the relevant usages are in `metrics_collector.go:133`, `metrics_collector.go:307`, `metrics_service.go:66`, and `oba_rest_api_metrics.go:74`.

Moreover, the PR is conflicting with the main branch consider resolve this conflict.

ahmadmuhamadi:
Okay, Tmrw i will work on them

aaronbrethorst:
### Code review

Found 1 issue:

1. The explanatory comment on the stub handler in the new test is truncated mid-sentence — "If any agency refetched instead of reading the threaded response, it would see" ends with no object and no period, so the reason the `n > 1` branch returns an empty `agencyIDs` list never gets stated. This looks like the leftover of the earlier review's request to finish the test comments (the function-level comment above it was updated; this one was not).

https://github.com/OneBusAway/watchdog/blob/fa10293cdf2e02fd89ad99715ccb2edce0dbe2f5/internal/app/metrics_collector_test.go#L143-L151

The prefetch logic itself checks out: `/api/where/metrics.json` is genuinely per-server and every per-agency value in the response is read out of a map keyed by `agencyID`, so sharing one decoded body across a server's agencies can't misattribute anything. `entry := cacheMetrics.Data.Entry` is a struct copy and nothing downstream mutates the shared maps or slices. A failed probe leaves `liveAgencies` nil, which empties `liveAgencyEntries` and returns before the per-agency loop, so server-mode never reaches `fetchObaAPIMetrics` with a nil `cacheMetrics` — there is no silent-degradation path. Backoff and ping failures short-circuit `collectAgencyChecks` before the prefetch is consumed, which is correct. The new test is not vacuous: it requires both agencies live, so the old behavior would have produced 3 requests against its `got != 1` assertion.

aaronbrethorst (changes requested):
This is a good change and the hard parts are right. Things I verified rather than assumed:

- **The scoping premise holds.** `/api/where/metrics.json` takes no agency parameter, and every per-agency value in `OBAMetrics.Data.Entry` — `RealtimeRecordsTotal`, `RealtimeTripCounts{Matched,Unmatched}`, `ScheduledTripsCount`, `StopIDs{Matched,Unmatched}Count`, `StopIDsUnmatched`, `TimeSinceLastRealtimeUpdate` — is a `map[string]...` indexed by `agencyID`. Nothing is read positionally, so one body genuinely does serve every agency on the server. That's the thing that had to be true for this PR to be safe, and it is.
- **No aliasing hazard.** `entry := cacheMetrics.Data.Entry` copies the struct, the maps inside are only indexed and ranged, and the one slice that escapes — `entry.StopIDsUnmatched[agencyID]` — goes to `GetStopLocationsByIDs`, which copies into a set and never sorts or mutates in place. I checked that specifically, since sharing one decode across agencies is new.
- **A failed prefetch can't silently degrade anything.** `probeLiveAgencies` returns `(nil, nil, err)`, `liveAgencies` is then a nil map, `liveAgencyEntries` comes out empty, and the pre-existing `if len(liveAgencyEntries) == 0 { return }` fires *before* the per-agency loop. So `collectAgencyChecks` is never reached with a nil `cacheMetrics` in server-mode, and that tick emits no agency metrics either way — exactly as it does on `main`. The nil branch is agency-mode-only in practice.
- **Backoff and ping still gate correctly.** `collectAgencyChecks` returns on backoff and on a failed `ServerPing` before `FetchObaAPIMetrics`, so a backing-off agency simply doesn't consume the prefetch. It's a plain value, so nothing leaks.
- **The test earns its keep.** Poisoning the second response and asserting both `metricsCalls == 1` and the distinct per-agency gauge values is the right construction — the old code would make three requests and fail that assertion. `go vet`, the full suite, and `-race` on `internal/app` and `internal/metrics` are all clean for me on the rebased head, and the conflict with `main` is resolved.

Two things before it lands. Both are leftovers from @0xaboomar's review that were only half-applied, and neither is a logic problem.

## 1. The comment in the new test is cut off mid-sentence

`internal/app/metrics_collector_test.go`, just above the `n > 1` branch:

```go
// Only the first response carries real data. If any agency
// refetched instead of reading the threaded response, it would see
if n > 1 {
```

It stops at "it would see" — no object, no period. That branch is the cleverest part of the test and the comment is what explains it, so it's worth finishing: something like "...it would see an empty `agencyIDs` list and record nothing, failing the gauge assertions below."

Worth noting nothing will catch this for you — gofmt, vet and CI are all happy with a truncated comment.

## 2. The function summary still says it always performs an HTTP GET

`internal/metrics/oba_rest_api_metrics.go`, the opening paragraph of `fetchObaAPIMetrics`:

> It performs an HTTP GET request to the server's `/metrics.json` endpoint using the provided API key, decodes the response into structured fields, and populates per-agency Prometheus metrics...

You correctly updated the `client` and `cacheMetrics` bullets further down, so the comment now contradicts itself within a dozen lines. A reader who stops at the summary — which is the one most people read — gets the pre-PR behavior. Reword it to say it populates the per-agency metrics from a response it either fetches itself or receives via `cacheMetrics`.

While you're in there: the `//` separator line before `Returns:` got dropped, so godoc renders `Returns:` as a continuation of the `cacheMetrics` bullet rather than its own section.

On @0xaboomar's optional rename: `cacheMetrics` is fine and I wouldn't churn it now, though `prefetchedMetrics` would have read better — it's a per-tick local, not a cache with a lifetime.

Fix those two and I'll merge it promptly. Nice optimization — this is a clean N+1 to 1 reduction with no behavioral risk, which is the best kind.

ahmadmuhamadi:
Hello @aaronbrethorst , 

I did the docs and resolved the conflicts with new updates in the main if there's any issue lmk, Ofc i will be happy to work on it 

Thanks :)

0xaboomar:
Hey @ahmadmuhamadi, looks good now. Previous feedback is addressed and CI is passing. Ready to merge!

